PR #27, opened 2022-05-16
bugfix: save node id in sequence when dag walking and node executed

howieyuen:
When `kusion apply` in preview stage, the previewed table not always the same.

## Root Cause

- after pre-dependent nodes finished, these downEdges are executing in paralyzing, so we cannot ensure the key returned from resource_node.execute() are alway the same.
- map traversing(node id -> action), it is originally disordered.

## Current Fix

- ignore concurrent execution of peer nodes
- save node id in sequence, keep order by traversing slice instead of map


elliotxx (approved):
lgtm

PR #57, opened 2022-06-02
engine: add some comments

SparkYuan:
1. re-organization engine pkgs
2. add some comments in the pkg engine to make codes easier to read

### DIFF
--- a/pkg/engine/operation/models/doc.go
+++ b/pkg/engine/operation/models/doc.go
@@ -1,0 +1,3 @@
+// Package models contains internal models of operations
+// todo CLI imports this package directly. We need to make this pkg internal
+package models


SparkYuan:
I have read the CLA Document and I hereby sign the CLA

pkg/engine/operation/apply.go
```diff
@@ -44,6 +44,11 @@ func NewApplyGraph(m *models.Spec, priorState *states.State) (*dag.AcyclicGraph,
 	return graph, s
 }
 
+// Apply means turn all actual infra resources into the desired state described in the request by invoking a specified Runtime.
+// Like other operations, Apply has 3 main steps during the whole process.
+// 1. parse resources and their relationship to build a DAG and should take care of those resources that will be deleted
+// 2. walk this DAG and execute all graph nodes concurrently, besides the entire process should follow dependencies in this DAG
+// 3. during the execution of each node, it will invoke different runtime according to the resource type
```
chai2010, line 51: the go doc do not support markdown format, the list item should split by blank line.
see https://pkg.go.dev/fmt

SparkYuan: these lines are just nomarl comments for the codes below....

pkg/engine/doc.go
```diff
@@ -0,0 +1,12 @@
...
+//
+// Kusion Engine is a middle layer between Spec and the actual infrastructure.
+// The major function of this engine is to parse the KCL compiled result and to turn all actual infra resources into the desired state
+// described in KCL. It consists of 3 parts:
+// 1. DAG engine: parse resources in the Spec and figure out which resource should be created,
+// updated or deleted. The analysis result will be converted into a DAG.
+// 2. Runtime: an interface that represents an actual infrastructure.
+// All operations that trying to manipulate a resource should delegate to one Runtime
+// 3. State: state is a record of an operation's result.
+// It is often used as a datasource for 3-way merge/diff in operations like Apply or Preview
```
ldxdl, line 11: + Kusion Engine is a middle layer between API spec and the actual infrastructure. And Who will use the Engine?
+ Remove 'KCL' in the doc.  API spec does not need to perceive the upper layer input.
+ API Spec should be versioned, and should be backwards compatible?
+ The conversion mechanism, codec, convention.. from api spec to internal model?
+ I prefer to split 'DAG engine' to a diff engine, a DAG builder and an execution engine?
+ Runtime: runtime is usually not an interface 
+ Important interface, structs and usage examples?

ldxdl: Judging from the code I have learned so far and his positioning, maybe this code structure would be more appropriate:

/types -- engine‘s internal types
/models -- engine's internal model
/exec -- nodes exec flow
/dag -- dag builder, walker, etc
/diff -- diff tools
/runtime -- runtime apis and impls
/state -- state apis and impls
api.go -- engine's core public interfaces and functionalities.
types.go -- entine's public types.

Through such code organization, the core functions of the engine can be better showed, and the codes of different functional modules can also be reasonably organized.

You can modify it as you like. Hope the advice helps.

pkg/engine/operation/apply.go
elliotxx: Comment need to follow official specifications:
https://go.dev/doc/comment

pkg/kusionctl/cmd/apply/options.go
```diff
@@ -8,12 +8,16 @@ import (
 	"strings"
 	"sync"
 
+	"kusionstack.io/kusion/pkg/engine/operation"
+	opsmodels "kusionstack.io/kusion/pkg/engine/operation/models"
```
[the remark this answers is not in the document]
SparkYuan: ok, I have added a doc file in pkg/engine/operation/models and I will make this pkg internal later

elliotxx (approved):
lgtm

PR #62, opened 2022-06-07
feat: support custom writer and storage for apply and preview

elliotxx:
Purpose:
* Public apply and preview func for external system integration
* Support custom state storage and writer for apply and preview func
* Bump the `github.com/pterm/pterm` package to the latest version


go.mod
```diff
@@ -41,7 +41,7 @@ require (
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
 	github.com/mitchellh/hashstructure v1.0.0
 	github.com/pkg/errors v0.9.1
-	github.com/pterm/pterm v0.12.41
+	github.com/pterm/pterm v0.12.42-0.20220427210824-6bb8c6e6cc77
```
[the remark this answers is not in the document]
elliotxx: The PR use unrelease features of `pterm`

howieyuen (approved):

howieyuen (approved):

pkg/kusionctl/cmd/apply/options.go
```diff
@@ -129,8 +131,14 @@ func (o *ApplyOptions) Run() error {
 	return nil
 }
 
-func preview(o *ApplyOptions, planResources *models.Spec,
-	project *projectstack.Project, stack *projectstack.Stack,
+// The preview function will preview for all resources changes
+func Preview(
+	o *ApplyOptions,
+	storage states.StateStorage,
```
ldxdl, line 137: argument not used

pkg/kusionctl/cmd/apply/options.go
```diff
@@ -129,8 +131,14 @@ func (o *ApplyOptions) Run() error {
 	return nil
 }
 
-func preview(o *ApplyOptions, planResources *models.Spec,
-	project *projectstack.Project, stack *projectstack.Stack,
+// The preview function will preview for all resources changes
```
ldxdl, line 134: Write full docstring for public apis. See
https://github.com/KusionStack/KCLVM/blob/eb32a3d87eac48450fa4fb6331470cb735c1cf44/kclvm/tools/src/query/mod.rs#L21
https://github.com/KusionStack/KCLVM/blob/eb32a3d87eac48450fa4fb6331470cb735c1cf44/kclvm/tools/src/query/mod.rs#L21

pkg/kusionctl/cmd/apply/options.go
```diff
@@ -166,7 +174,15 @@ func preview(o *ApplyOptions, planResources *models.Spec,
 	return operation.NewChanges(project, stack, rsp.Order), nil
 }
 
-func apply(o *ApplyOptions, planResources *models.Spec, changes *operation.Changes) error {
+// The apply function will apply the resources changes,
+// and will save the state to specified storage.
+func Apply(
```
ldxdl, line 179: Check your pub apis, you'd better not write too many details in the pub api directly, try to keep the core logic/flow clear and readable, and preferably no more than one screen in length.

pkg/kusionctl/cmd/apply/options.go
elliotxx: nice catch, fixed

healthjyk (commented):
look good to me

howieyuen (approved):

ldxdl:
The interface design of the engine and the options’ pub API is not very reasonable. As here, the runtime interface shouldn't be exposed. @elliotxx and @SparkYuan, pls pay more attention to design and work out a more reasonable API interfaces of option and engine layer. And, we need to reconsider the position and functionality of the option layer, cause each individual layer has its design and cost. 

BTW. In general, the interface parameters include three parts, the configurations(str, int, bool..), the feature flags(bools..) and the functional structures(IO, client, etc). The functional structures are injected into the lower layer by the middle layer, and obtained by the factory provided by the lower layer. We can design and implement APIs more standardly.

elliotxx:
@ldxdl OK, the interface design will continue to be optimized, and we will submit PR separately to solve.

ldxdl (approved):
LGTM

PR #64, opened 2022-06-07
bugfix: runtime apply and detail prompt cancel unexpectly

howieyuen:
### issue

fix: https://github.com/KusionStack/kusion/issues/65

### action
- runtime apply should differ create or patch, json merge patch cannot applied to an object which is not existed
- prompt cycle has a cancel condition, which is triggered every time
- refactor computing node's action logic
- use live state and plan state to make a diff report which is consistent with runtime logic

### DIFF
--- a/pkg/engine/operation/preview_test.go
+++ b/pkg/engine/operation/preview_test.go
@@ -29,6 +29,10 @@ var (
 		ID:         "fake-id",
 		Attributes: FakeService,
 	}
+	FakeResourceState2 = models.Resource{
+		ID:         "fake-id-2",
+		Attributes: FakeService,
+	}
 )
 
 var _ runtime.Runtime = (*fakePreviewRuntime)(nil)
@@ -40,6 +44,9 @@ func (f *fakePreviewRuntime) Apply(ctx context.Context, priorState, planState *m
 }
 
 func (f *fakePreviewRuntime) Read(ctx context.Context, resourceState *models.Resource) (*models.Resource, status.Status) {
+	if resourceState.ResourceKey() == "fake-id" {
+		return nil, nil
+	}
 	return resourceState, nil
 }
 
@@ -65,8 +72,7 @@ func TestOperation_Preview(t *testing.T) {
 		lock                    *sync.Mutex
 	}
 	type args struct {
-		request   *PreviewRequest
-		operation Type
+		request *PreviewRequest
 	}
 	tests := []struct {
 		name    string
@@ -78,9 +84,10 @@ func TestOperation_Preview(t *testing.T) {
 		{
 			name: "success-when-apply",
 			fields: fields{
-				Runtime:      &fakePreviewRuntime{},
-				StateStorage: &states.FileSystemState{Path: states.KusionState},
-				Order:        &ChangeOrder{StepKeys: []string{}, ChangeSteps: map[string]*ChangeStep{}},
+				OperationType: ApplyPreview,
+				Runtime:       &fakePreviewRuntime{},
+				StateStorage:  &states.FileSystemState{Path: states.KusionState},
+				Order:         &ChangeOrder{StepKeys: []string{}, ChangeSteps: map[string]*ChangeStep{}},
 			},
 			args: args{
 				request: &PreviewRequest{
@@ -96,7 +103,6 @@ func TestOperation_Preview(t *testing.T) {
 						},
 					},
 				},
-				operation: Apply,
 			},
 			wantRsp: &PreviewResponse{
 				Order: &ChangeOrder{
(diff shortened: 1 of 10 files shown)


pkg/engine/operation/resource_node.go
```diff
@@ -23,59 +23,67 @@ var _ ExecutableNode = (*ResourceNode)(nil)
...
 	if status.IsErr(s) {
 		return s
 	}
 
-	// 3. compute ActionType of current resource node between planState and liveState
-	if liveState == nil {
-		rn.Action = Create
-	} else if planedState == nil {
+	// 4. compute ActionType of current resource node between planState and liveState
+	switch operation.OperationType {
```
SparkYuan, line 50: Isn't it enough with comparing betweent states to determinate the resource's action? Why we should import OperationType in resource_node?

pkg/engine/operation/resource_node.go
```diff
@@ -23,59 +23,67 @@ var _ ExecutableNode = (*ResourceNode)(nil)
...
-	if rn.Action != Delete {
-		if operation.OperationType != Preview {
-			// replace implicit references
-			value := reflect.ValueOf(rn.state.Attributes)
-			_, implicitValue, s := ParseImplicitRef(value, operation.CtxResourceIndex, ImplicitReplaceFun)
-			if status.IsErr(s) {
-				return s
-			}
-			rn.state.Attributes = implicitValue.Interface().(map[string]interface{})
+	if operation.OperationType == Apply {
```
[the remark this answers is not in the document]
howieyuen: ok, I'll change the logic back to non-preview operation

pkg/engine/operation/resource_node.go
howieyuen: I have these thoughts:
1、we have two operations which will make changes on resource, one is apply, other is destroy
2、destroy operation doesn't need compare, it must be Delete action 
3、apply need compare, cause kusion apply contains create/update/delete action

pkg/engine/operation/types.go
```diff
@@ -6,6 +6,7 @@ type Type int64
 const (
 	UndefinedOperation Type = iota // invalidate value
 	Apply
-	Preview
+	ApplyPreview
```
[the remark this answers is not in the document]
howieyuen: preview is used in Apply and Destroy, which is judged with Operation type, aka:
- apply cmd + preview: need to compute action
- destroy cmd + preview: must be delete

here, I split it into 2 kind, make two meanings:
1. preview or make change
2. apply or delete

pkg/engine/operation/resource_node.go
```diff
@@ -23,59 +23,67 @@ var _ ExecutableNode = (*ResourceNode)(nil)
...
-	if rn.Action != Delete {
-		if operation.OperationType != Preview {
-			// replace implicit references
-			value := reflect.ValueOf(rn.state.Attributes)
-			_, implicitValue, s := ParseImplicitRef(value, operation.CtxResourceIndex, ImplicitReplaceFun)
-			if status.IsErr(s) {
-				return s
-			}
-			rn.state.Attributes = implicitValue.Interface().(map[string]interface{})
+	if operation.OperationType != DestroyPreview && operation.OperationType != Destroy {
```
SparkYuan, line 26: applyPreview also don't need to replace implicit references according to my knowledge

howieyuen: ok...that means only `Apply` need this

pkg/engine/operation/preview_test.go
```diff
@@ -40,7 +44,11 @@ func (f *fakePreviewRuntime) Apply(ctx context.Context, priorState, planState *m
 }
 
 func (f *fakePreviewRuntime) Read(ctx context.Context, resourceState *models.Resource) (*models.Resource, status.Status) {
-	return resourceState, nil
+	if resourceState.ResourceKey() == "fake-id" {
+		return nil, nil
+	} else {
+		return resourceState, nil
+	}
```
elliotxx, line 51: 简写为：
```
if resourceState.ResourceKey() == "fake-id" {
    return nil, nil
}

return resourceState, nil
```

pkg/engine/runtime/kubernetest_runtime.go
```diff
@@ -57,34 +58,41 @@ func (k *KubernetesRuntime) Apply(ctx context.Context, priorState, planState *mo
...
+		// modified equals input content
+		modified := json.MustMarshal2String(planState.Attributes)
+		// current equals live manifest
+		current := json.MustMarshal2String(liveState.Attributes)
+		// 3-way json merge patch
+		patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch([]byte(original), []byte(modified), []byte(current))
+		if err != nil {
+			return nil, status.NewErrorStatus(err)
+		}
+		// apply patch
```
elliotxx, line 87: 注释建议统一为首字母大写

SparkYuan (approved):
lgtm

PR #70, opened 2022-06-09
fix unit test for pkg/scaffold/internal

howieyuen:
### TestTemplateRepository_Templates
UT `read from subdir` means to fetch 2 or more templates from `pkg/scaffold/internal`, now we have only one template, which is not compatible with more templates in the future.

### TestCopyTemplateFiles

The flaw of this one is pretty the same as above. Copy templates from src dir to target dir, now src dir contains only one, so it's not compatible with more templates as well.

### DIFF
--- a/pkg/scaffold/external_templates_test.go
+++ b/pkg/scaffold/external_templates_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"bou.ke/monkey"
+	"github.com/jinzhu/copier"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/gitutil"
 	"github.com/stretchr/testify/assert"
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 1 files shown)


elliotxx:
Need to supplement PR description

pkg/scaffold/external_templates_test.go
```diff
@@ -266,8 +264,9 @@ func TestCopyTemplateFiles(t *testing.T) {
 		}
 		stack2Configs[stack.Name] = configs
 	}
-	// project name
-	projectName := localTemplate.ProjectName
-	err = CopyTemplateFiles(localRoot, tmp, true, projectName, projectConfigs, stack2Configs)
+	err = CopyTemplateFiles(localTemplate.Dir, tmp, true, localTemplate.ProjectName, projectConfigs, stack2Configs)
+	if err != nil {
+		fmt.Printf("err: %v", err)
```
elliotxx, line 269: It is recommended not to use `fmt.Printf` output error directly in a unit test

howieyuen: sorry...this one is used to debug error...

elliotxx (approved):
lgtm

markliby (approved):
lgtm

Issue #72, opened 2022-06-13
`kusion init` provides  a new way to collect template params value with specified CLI flag


howieyuen:
### Summary

`kusion init` is a helpful tool to create a new project of KCL. Consider that we have two different roles, one is **template developer**, who is responsible for design and develop template, as well as a set of default values; another is to use template. For quick start of template rendering, developer provides a YAML file, named *kusion.yaml*, which contains type and default value of all params that is used by go template engine.

However, if I want to override template default value, there is only one solution: type new default when each **Prompt** asks, which is strongly Inefficient.

### Proposal

Add a new flag(CustomParams) to override template defaults. Kusion must to provide a struct with filed name and type. User can prepare a new set of defaults in advance, based on file *kusion.yaml*.

#### Field Types

Although we encourage all to use primitive type for template params:
- string
- int
- float
- bool

But we must admit there is always an exception, aka complicated data type, For example:
- array
- map
- struct
- any
 
and mixed complicated with primitives. So we need to explain every field type in *kusion.yaml*.

#### Flag Definition in Struct

*kusion.yaml* provide a struct as below: 

<details>
<summary>kusion.yaml</summary>

```yaml
projectName: my-app
description: A minimal kusion project of single stack
quickstart: kusion compile main.k -Y ci-test/settings.yaml
common:
  # base/base.k
  - name: ServiceName
    description: service name
    type: string
    default: frontend-svc
  - name: NodePort
    description: node port
    type: int
    default: 30000
  # project.yaml
  - name: ProjectName
    description: project name
    type: string
    default: my-app
stacks:
  - name: dev
    fields:
      # stack/ci-test/settings.yaml
      - name: Stack
        description: stack env. One of dev,test,stable,pre,sim,gray,prod.
        type: string
        default: dev
      # stack/main.k
      - name: Image
        description: The Image Address. Default to 'gcr.io/google-samples/gb-frontend:v4'
        type: string
        default: gcr.io/google-samples/gb-frontend:v4
      # stack/stack.yaml
      - name: ClusterName
        description: The Cluster Name. Default to 'kubernetes-dev'
        type: string
        default: kubernetes-dev
```

</details>

Implicit struct of CustomParams flag:
<details>
<summary>struct definition</summary>

```json
{
	"projectConfig": {
		"ServiceName": "frontend-svc",
		"NodePort": 30000,
		"ProjectName": "my-app"
	},
	"stacksConfig": {
		"dev": {
			"Stack": "dev",
[...]

PR #73, opened 2022-06-14
add new FiledType: MapField and AnyField to indicate map entry and interface{}

howieyuen:
ref: https://github.com/KusionStack/kusion/issues/72

This PR mainly to add missing field type for template render engine, especially composite type, eg: array, map and struct.

And I also rename struct field of ProjectTemplate, make it more accurate and consistency with custom params structure.


elliotxx (approved):
lgtm

markliby (approved):

PR #78, opened 2022-06-22
feat: refactor CopyTemplateFiles to support different file system

howieyuen:
related: https://github.com/KusionStack/kusion/issues/72#issuecomment-1157618380

Refactor `CopyTemplateFiles()`(rename to `RenderLocalTemplate()`) to support different file system, not only local directory:
1. For the built-in templates, we must use the `go:embed` command, otherwise, we cannot compile it into binary, and we cannot access the built-in templates directory because of a relative path.
2. Render the template to the specified directory, the whole process is divided into 3 steps:
   1. `ReadTemplate()`: read the local path to srcFS
   2. `RenderFSTemplate()`: render srcFS to destFS
   3. `WriteToDisk()`: write to disk
3. For third-party integration scenarios, you can choose one:
   1. Local directory: `RenderLocalTemplate()`
   2. Self-provided memory file system: `RenderFSTemplate()`
   3. Built-in template that comes with the scaffold package: `GetInternalTemplates()`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pterm/pterm v0.12.42-0.20220427210824-6bb8c6e6cc77
 	github.com/sergi/go-diff v1.2.0
+	github.com/spf13/afero v1.2.2
 	github.com/spf13/cobra v1.1.1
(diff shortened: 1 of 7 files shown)


elliotxx (approved):
lgtm

Issue #75, opened 2022-06-17
Concurrent execute `kusion compile` error


elliotxx:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

After concurrent execute `kusion compile` for a period of time.

### 2. What did you expect to see? (Required)

Normal compilation without error

### 3. What did you see instead (Required)

![image](https://user-images.githubusercontent.com/9360247/174216181-96ac5a33-928a-4ec3-b85d-e9d809909339.png)

### 4. What is your KusionStack components version? (Required)

```
releaseVersion: v0.4.2-alpha.1
gitInfo:
    latestTag: v0.4.2-alpha.1
    commit: 28b4f935cdd2813f5e7c09de855f6e584ed4af3c
    treeState: clean
buildInfo:
    goVersion: go1.17.10
    GOOS: linux
    GOARCH: amd64
    numCPU: 2
    compiler: gc
    buildTime: "2022-05-26 08:37:19"
dependency:
    kclvmgoVersion: v0.4.2-alpha4
    kclPluginVersion: v0.4.1-alpha2
```

elliotxx:
We can add a `--silent-log` flag of `kusion compile` to silent log output to file

PR #82, opened 2022-06-27
feat: util/log support log to stderr and add log-to-stderr flag for kusion compile

howieyuen:
fix: https://github.com/KusionStack/kusion/issues/75

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -2,6 +2,7 @@ package log
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"strings"
 )
@@ -97,4 +98,8 @@ func GetLogger() Logger {
 
 func With(args ...interface{}) Logger {
 	return log.With(args...)
+}
+
+func SetOutput(out io.Writer) {
+	log.SetOutput(out)
 }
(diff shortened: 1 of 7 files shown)


elliotxx (approved):
lgtm

pkg/log/log.go
```diff
@@ -98,3 +98,8 @@ func GetLogger() Logger {
 func With(args ...interface{}) Logger {
 	return log.With(args...)
 }
+
+// LogToStderr sets whether log to stderr instead of files
+func LogToStderr(silent bool) {
```
elliotxx, line 103: slient is typo?

pkg/log/log.go
```diff
@@ -98,3 +98,8 @@ func GetLogger() Logger {
...
+	log.LogToStderr(silent)
+}
```
elliotxx, line 105: Consider:
```
func SetOutput(out io.Writer) {
	log.SetOutput(out)
}
```

Also see: https://github.com/sirupsen/logrus/blob/master/logger.go#L391

pkg/log/log.go
howieyuen: sorry...

pkg/log/log.go
howieyuen: Good Idea

elliotxx (approved):
lgtm

markliby (approved):

PR #105, opened 2022-07-23
bugfix: stack directory name is map key of Template.StacksConfig

howieyuen:
wrong use `stack.yaml` as map key of Template.StackConfigs, stack directory name is correct.


markliby (approved):

elliotxx (approved):
lgtm

Issue #102, opened 2022-07-18
Missing resources in the kusion_state.json file when applying resources already in the k8s cluster


SparkYuan:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. Resource A is described in a KCL file
2. Resource A is already in the actual k8s cluster
3. When I apply this resource with command `apply`, Kusion will recognize this resoruce as `Unchanged` and will `Skip` this resource.

### 2. What did you expect to see? (Required)
Resource A is saved in kusion_state.json file

### 3. What did you see instead (Required)
Resource A isn't saved in kusion_state.json file

### 4. What is your KusionStack components version? (Required)
v0.5.0

PR #107, opened 2022-07-26
bugfix: save unchanged but lived resource

howieyuen:
fix: https://github.com/KusionStack/kusion/issues/102

`Resource` is living in `Runtime`, but not applied by kusion, next time user use kusion to apply the same `Resource` but no changes, kusion should save it in `State`


elliotxx (approved):
lgtm

markliby (approved):
lgtm

PR #108, opened 2022-07-26
bugfix: resource nodes of delete action don't need compute action type again

howieyuen:
fix: https://github.com/KusionStack/kusion/issues/109

**Root Causes**

- `kusion apply` will build an acyclic graph which contains  only two kinds of action, `Update` and `Delete`, only `Update` nodes need to determine `Create` or `Update`
- Nodes of `Update` action, its state field is **plan state**; otherwise, node of `Delete` action, its state field is **prior state**
- For nodes of `Delete` action, wrongly used `node.state` as plan state in `execute()` func, so `Delete` turns to `Unchange`


elliotxx (approved):
lgtm

markliby (approved):
lgtm

Issue #110, opened 2022-08-03
invalid argument for "-D, --argument" flag: parse error, bare " in non-quoted-field


howieyuen:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

kusion compile a.k -D a='["x","y","z"]' -D b='{"foo": "bar"}'

```kcl
a = option("a")
b = option("b")
```

### 2. What did you expect to see? (Required)

```yaml
a:
- x
- y
- z
b:
  foo: bar
```

### 3. What did you see instead (Required)

```shell
 ❌  invalid argument "a=[\"x\",\"y\",\"z\"]" for "-D, --argument" flag: parse error on line 1, column 4: bare " in non-quoted-field
      See 'kusion compile --help' for usage.
```

### 4. What is your KusionStack components version? (Required)

```shell
releaseVersion: v0.5.2
```

PR #111, opened 2022-08-03
bugfix: invalid argument for "-D, --argument" flag: parse error, bare " in non-quoted-field

howieyuen:
fix: #110 

kusion compile flag,  `-D` use `pflag.StringSliceVarP()` to parse string slice arguments, according to the comments here:
https://github.com/spf13/pflag/blob/d5e0c0615acee7028e1e2740a11102313be88de1/string_slice.go#L99
```
compared to StringArray flags, StringSlice flags take comma-separated value as arguments and split them accordingly.
```
so, use `pflag.StringArrayVarP()` instead, will prevent pflag from truncating long string by comma.

After replace, the compiled result meets expectation:
```bash
./kusion compile a.k -D a='["x","y","z"]' -D b='{"key": False}'
```

output:
```yaml
a:
- x
- y
- z
b:
  key: false
```

### DIFF
--- a/pkg/kusionctl/cmd/compile/compile.go
+++ b/pkg/kusionctl/cmd/compile/compile.go
@@ -56,20 +56,24 @@ func NewCmdCompile() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringSliceVarP(&o.Arguments, "argument", "D", []string{},
-		i18n.T("Specify the top-level argument"))
-	cmd.Flags().StringSliceVarP(&o.Settings, "setting", "Y", []string{},
-		i18n.T("Specify the command line setting files"))
+	o.AddCompileFlags(cmd)
 	cmd.Flags().StringVarP(&o.Output, "output", "o", "",
 		i18n.T("Specify the output file"))
-	cmd.Flags().StringVarP(&o.WorkDir, "workdir", "w", "",
-		i18n.T("Specify the work directory"))
 	cmd.Flags().BoolVarP(&o.DisableNone, "disable-none", "n", false,
 		i18n.T("Disable dumping None values"))
 	cmd.Flags().BoolVarP(&o.OverrideAST, "override-AST", "a", false,
 		i18n.T("Specify the override option"))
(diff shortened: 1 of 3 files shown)


pkg/kusionctl/cmd/compile/compile.go
```diff
@@ -56,20 +56,24 @@ func NewCmdCompile() *cobra.Command {
...
+func AddCompileFlags(cmd *cobra.Command, o *CompileOptions) {
+	cmd.Flags().StringVarP(&o.WorkDir, "workdir", "w", "",
+		i18n.T("Specify the work directory"))
+	cmd.Flags().StringSliceVarP(&o.Settings, "setting", "Y", []string{},
+		i18n.T("Specify the command line setting files"))
+	cmd.Flags().StringArrayVarP(&o.Arguments, "argument", "D", []string{},
+		i18n.T("Specify the top-level argument"))
+	cmd.Flags().StringSliceVarP(&o.Overrides, "overrides", "O", []string{},
+		i18n.T("Specify the configuration override path and value"))
+}
```
elliotxx, line 79: Consider adding a func to the XXXXOptions:
```go
func (o *CompileOptions) AddFlags(cmd *cobra.Command) {
	cmd.Flags().StringVarP(&o.WorkDir, "workdir", "w", "",
		i18n.T("Specify the work directory"))
	cmd.Flags().StringSliceVarP(&o.Settings, "setting", "Y", []string{},
		i18n.T("Specify the command line setting files"))
	cmd.Flags().StringArrayVarP(&o.Arguments, "argument", "D", []string{},
		i18n.T("Specify the top-level argument"))
	cmd.Flags().StringSliceVarP(&o.Overrides, "overrides", "O", []string{},
		i18n.T("Specify the configuration override path and value"))
}
```
Can refer: https://github.com/kubernetes/component-base/blob/85b8d4a6aaaaf51bbd0dbcf732bc843448abedd0/featuregate/feature_gate.go#L317

howieyuen: good advice~

elliotxx (approved):
lgtm

markliby (approved):
lgtm

PR #132, opened 2022-08-26
feat: kusion apply support ServerSide preview

howieyuen:
fix: https://github.com/KusionStack/kusion/issues/133

Kusion will try `ServerSideDryRun` first, if runtime doest not support `ServerSideDryRun`, kusion will fall back to `ClientSideDryRun`, user will not perceive it.


markliby (approved):
lgtm

Issue #23, opened 2022-05-12
Feat: add watch subcommand to observe specified resource lifecycle


howieyuen:
## Background

The initial requirement of `kusion watch` comes from the fact that after the user successfully creates, modifies or deletes the resource using the `kusion apply`, which can monitor the real-time update of the resource, so as to ensure that Spec and Live are ultimately consistent on the Runtime side. Therefore, the operation object of the Watch subcommand is not a single resource, but a single Stack under a certain Project.

Kusion has been connected to Kubernetes and Terraform. For k8s resources, client-go provides Watch interface, resource changes that can be obtained in real time; and for the Terraform ecosystem, there is no so-called Live Manifest, and the Terraform side just saves The resource list after the operation is successfully executed, which is equivalent to a cache.

## Target

As an alpha feature, the primary purpose is to define the watch workflow. So TF runtime is not include in this time. And there are a few preconditions:
- only for k8s
- only for namespace, service and deployment

## Watch Workflow

The complete workflow can be simplified as 4 stages:

1. Compile Project&Stack to get all resources in Stack
2. Call the IaC Engine interface to start watching
3. IaC Engine asynchronously triggers the Runtime Watch interface for the resource list
4. Runtime implements Watch logic and customizes the prompt Message

```mermaid
sequenceDiagram
    autonumber
    actor User
    participant KusionCLI
    participant KCLVM
    participant IaC Engine
    participant Runtime
    User ->> KusionCLI: kusion apply --watch
    KusionCLI->>KCLVM: compile
    activate KCLVM
    KCLVM-->>KusionCLI: kcl resoults
    deactivate KCLVM
    KusionCLI ->> IaC Engine: watch resources
    activate IaC Engine
    loop Every Resource
    IaC Engine ->> Runtime: watch one
    activate Runtime
    Runtime -->> IaC Engine: event
    deactivate Runtime
    IaC Engine ->> IaC Engine: render table row
    end
    IaC Engine -->> KusionCLI: render table 
    deactivate IaC Engine
   KusionCLI --> User: flush table 
```

### kubectl

For k8s, the `--w` flag in the kubectl get command provides the implementation of related capabilities. The Watch behavior of k8s for resources is **unterminated**, and the object of Watch can only be one resource or one certain type of resource. And formatted output for each resource, defined here `pkg/printers/internalversion/printers.go`. The key implementation of `kubectl get -w` is as follows:

```go
[...]

PR #144, opened 2022-10-11
feat: kusion watch supports resources of corev1 kinds

howieyuen:
related: https://github.com/KusionStack/kusion/issues/23#issuecomment-1272855221


SparkYuan (approved):
lgtm

PR #187, opened 2022-12-05
fix: differentiate dry run action based on live state is nil or not

howieyuen:


pkg/engine/runtime/kubernetes_runtime.go
```diff
@@ -116,20 +117,21 @@ func (k *KubernetesRuntime) Apply(ctx context.Context, request *ApplyRequest) *A
...
-			// Fall back to ClientSideDryRun
-			mergedPatch, err := jsonpatch.MergePatch([]byte(current), patchBody)
-			if err != nil {
-				return &ApplyResponse{nil, status.NewErrorStatus(err)}
-			}
-			res = &unstructured.Unstructured{}
-			if err = res.UnmarshalJSON(mergedPatch); err != nil {
-				return &ApplyResponse{nil, status.NewErrorStatus(err)}
+				// Fall back to ClientSideDryRun
+				log.Errorf("ServerSideDryRun patch %s failed, fall back to ClientSideDryRun; err: %v", planState.ID, err)
```
SparkYuan, line 122: This err msg will be printed twice
Duplicated Code in  L108

pkg/engine/runtime/kubernetes_runtime.go
```diff
@@ -116,20 +117,21 @@ func (k *KubernetesRuntime) Apply(ctx context.Context, request *ApplyRequest) *A
...
+
+				// Merge 3-way patch
+				mergedPatch, err := jsonpatch.MergePatch([]byte(current), patchBody)
```
SparkYuan, line 125: Why do we need another merge here since `patchBody` already ThreeWayJSONMergePatch before in L91

pkg/engine/runtime/kubernetes_runtime.go
howieyuen: I am afraid not, there are two conditions, one is live state is nil, it a creating verb; another live state is not nil, its a patch verb.
they are totally different entering logic.

> This err msg will be printed twice Duplicated Code in L108

pkg/engine/runtime/kubernetes_runtime.go
howieyuen: > Why do we need another merge here since `patchBody` already ThreeWayJSONMergePatch before in L91

`ThreeWayJSONMergePatch()` will return a patch body, not a patch result.

SparkYuan (approved):
lgtm

markliby (approved):

Issue #149, opened 2022-10-21
Feature Request: Secret as Code


howieyuen:
## Feature Request

Let me introduces 2 scenes here:

1. some apps store ak/sk in there env variables
2. some apps need cert file which is stored in k8s secret

Since Konfig is a public repository, it is obviously inappropriate to store the sensitive information in the above scenarios in a large repository. Therefore, this issue explores solutions for maintaining sensitive information in a large repository.

howieyuen:
## SealedSecrets

SealSecrets is consist of:
- cluster controller: `sealed-secrets-controller`
- user CLI: `kubeseal`

`SealedSecret` object has a template section which encodes all the fields you want the controller to put in the unsealed `Secret`. This includes metadata such as labels or annotations, but also things like the type of the secret.
```yaml
apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  name: mysecret
  namespace: mynamespace
  annotations:
    "kubectl.kubernetes.io/last-applied-configuration": ....
spec:
  encryptedData:
    .dockerconfigjson: AgBy3i4OJSWK+PiTySYZZA9rO43cGDEq.....
  template:
    type: kubernetes.io/dockerconfigjson
    # this is an example of labels and annotations that will be added to the output secret
    metadata:
      labels:
        "jenkins.io/credentials-type": usernamePassword
      annotations:
        "jenkins.io/credentials-description": credentials from Kubernetes
```

After decrypting, generate a raw k8s secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
  namespace: mynamespace
  labels:
    "jenkins.io/credentials-type": usernamePassword
  annotations:
    "jenkins.io/credentials-description": credentials from Kubernetes
  ownerReferences:
  - apiVersion: bitnami.com/v1alpha1
    controller: true
    kind: SealedSecret
    name: mysecret
    uid: 5caff6a0-c9ac-11e9-881e-42010aac003e
type: kubernetes.io/dockerconfigjson
data:
  .dockerconfigjson: ewogICJjcmVk...
```
As you can see, the generated Secret resource is a "dependent object" of the SealedSecret and as such it will be updated and deleted whenever the SealedSecret object gets updated or deleted.

### Asymmetric Encryption

[...]

PR #202, opened 2022-12-26
feat: support vault as secret store

howieyuen:
related: https://github.com/KusionStack/kusion/issues/149#issuecomment-1311200340

project example: https://github.com/KusionStack/konfig/pull/77


pkg/vals/secret_stores.go
```diff
@@ -0,0 +1,37 @@
+package vals
```
SparkYuan, line 1: what does this word "vals“ short for? "values"? If this pkg is designed for SecretStores, maybe you can add a doc.go file or change it to a more understandable one

howieyuen: ok, I will add a doc.go

pkg/kusionctl/cmd/apply/options.go
```diff
@@ -150,6 +151,17 @@ func (o *ApplyOptions) Run() error {
 		}
 	}
 
+	// Replace ref secrets if needed
+	secure, ok := stack.Labels[projectstack.LabelSecure].(bool)
+	if ok && secure {
+		if !project.SecretStores.IsValid() {
+			return fmt.Errorf("no secret store is provided")
+		}
+		if err = vals.ReplaceRefs(planResources, project.SecretStores); err != nil {
```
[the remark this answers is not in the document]
howieyuen: I try to merge this logic into node execute(), it looks like not quite suitable, since secret ref need a secured label marking there is one or many secret refs need to be restore. 
And in the restoring logic, need secretStore definition, which is also required, so the func definition is changed, and a lot variables need to passed.

pkg/engine/operation/graph/resource_node.go
```diff
@@ -252,6 +254,11 @@ func ParseImplicitRef(v reflect.Value, resourceIndex map[string]*models.Resource
 			}
 			v = tv
 		}
+		if secured {
```
SparkYuan, line 257: although `secured` can optimize some speed but not necessary, because IsSecured() has already checked whether this resource contains encrypted values. 

Delete `secured` can also reduce an input params of this func...

howieyuen: No, you are wrong, `secured` comes from stack label, which means that value's format can be `ref+<secret store>`, but not secret ref, it's up to user, we cannot kidnap users

SparkYuan: Once `ref+<secret store>` is shown in user's code, that means this code is `secured`. Why should user add an extra config in stack label? Under what conditions users add `secured` label in stack and no secret ref in codes?

howieyuen: yes, I learn this from other product, eg: pulumi following this rule. Besides, secured label is not created by my own, I saw it in previous encryption solution.

markliby (approved):
lgtm

Issue #205, opened 2022-12-28
kusion compile -o option produces error


amyXia1994:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
```
git clone https://github.com/KusionStack/konfig
cd konfig
kusion compile -w appops/guestbook-frontend/pre -o output.yaml
```

### 2. What did you expect to see? (Required)

compiled result in `output.yaml` file

### 3. What did you see instead (Required)

```
 ❌  KCL Runtime Error[E3M38] : Evaluation failure
      ---> File /Users/amy/work/open/konfig/appops/guestbook-frontend/pre/main.k:5
      5 |appConfiguration: frontend.Server { -> Failure
      attribute 'mainContainer' of Server is required and can't be None or Undefined
```

### 4. What is your KusionStack components version? (Required)

```
kusion version
releaseVersion: v0.7.2
gitInfo:
    latestTag: v0.7.2
    commit: fb996130c1e4a8a30da7c04e672433ea6c870384
    treeState: clean
buildInfo:
    goVersion: go1.17.13
    GOOS: linux
    GOARCH: amd64
    numCPU: 2
    compiler: gc
    buildTime: "2022-12-02 07:00:28"
dependency:
    kclvmgoVersion: v0.4.4-alpha.1
    kclPluginVersion: v0.4.1-alpha2
```

howieyuen:
Thanks for reporting, this is a bug and the root cause as below:

Users can specify a stack dir as workDir, use flag `-w`, so kusion will preset `o.Settings` and `o.Output` which are default in `ci-test` directory. But the condition is extremely hard, need `o.Settings`, `o.Output` and `o.Filenames` all empty.

So I change a little, preset them separately according to each own condition.

PR #206, opened 2022-12-28
bugfix: preset o.Settings and o.Output separately

howieyuen:
close: https://github.com/KusionStack/kusion/issues/205

### DIFF
--- a/pkg/kusionctl/cmd/compile/options.go
+++ b/pkg/kusionctl/cmd/compile/options.go
@@ -94,8 +94,11 @@ func (o *CompileOptions) PreSet(preCheck func(cur string) bool) {
 		return
 	}
 
-	if len(o.Settings) == 0 && o.Output == "" && len(o.Filenames) == 0 {
+	if len(o.Settings) == 0 {
 		o.Settings = []string{filepath.Join(projectstack.CiTestDir, projectstack.SettingsFile), projectstack.KclFile}
+	}
+
+	if o.Output == "" {
 		o.Output = filepath.Join(projectstack.CiTestDir, projectstack.StdoutGoldenFile)
 	}
 }


markliby (approved):

pkg/kusionctl/cmd/compile/options.go
```diff
@@ -94,8 +94,11 @@ func (o *CompileOptions) PreSet(preCheck func(cur string) bool) {
 		return
 	}
 
-	if len(o.Settings) == 0 && o.Output == "" && len(o.Filenames) == 0 {
+	if len(o.Settings) == 0 {
```
amyXia1994, line 97: 1. should settings be set up to default value when the `o.Filenames` is not empty?
since the settings file may configs entrance file list, which are conflict with `o.Filenames`

2. should settings be set up to default value when the `o.Output` is not empty?
since the settings file may configs output file path, which is conflict with `o.Output`

howieyuen: the `settings.yaml` file can define entrance files, so do the cmd args, the o.Filenames is read from cmd args, not a flag.  and `settings.yaml` can also do the output file path ? really?

amyXia1994: about settings.yaml, you can see https://kusionstack.io/docs/reference/lang/lang/tour/#arguments-with-setting-files and checkout the kcl_cli_configs key

howieyuen: `kcl_cli_configs` is used in `kcl.yaml`, `kcl_options` is used in `settings.yaml`

amyXia1994: > `kcl_cli_configs` is used in `kcl.yaml`, `kcl_options` is used in `settings.yaml` for now

kcl.yaml and ci-test/settings.yaml are all settings files, only distinguish in names

amyXia1994: cc @Peefy

amyXia1994 (approved):

pkg/kusionctl/cmd/compile/options.go
howieyuen: accoding to @Peefy, user-specified filenames will override filenames in `kcl.yaml`, so it's is ok to compile

PR #222, opened 2023-01-13
auto import resources exist in spec and live cluster but no recorded in kusion_state.json

SparkYuan:
auto import resources exist in spec and live cluster but no recorded in kusion_state.json


markliby (approved):
lgtm

Issue #223, opened 2023-01-17
[Bug] There will be a `FileNotFound` error when installing kusion on some platforms.


Peefy:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Install kusion through `brew` or the `go install` command, and run:

```bash
echo 'a = 1' > a.k && kusion compile a.k
```

This problem seems to occur only on some machines (such as mac m1 and CI). For more error information, see:

+ brew install: https://github.com/KusionStack/kcl-lang.io/actions/runs/3936706315/jobs/6735820439
+ go install: https://github.com/KusionStack/kcl-lang.io/actions/runs/3936706315/jobs/6735821197

However, kclvm-go can be installed and run successfully through 'go install', see:

+ https://github.com/KusionStack/kcl-lang.io/actions/runs/3945431794

### 2. What did you expect to see? (Required)

The YAML output

```yaml
a: 1
```

### 3. What did you see instead (Required)

The error message

```stderr
 ❌  open : no such file or directory
```

### 4. What is your KusionStack components version? (Required)

```
releaseVersion: v0.7.3
gitInfo:
    latestTag: v0.7.3
    commit: 331a4b26eb325475f4dfd92d135f4f16f0111fdd
    treeState: clean
buildInfo:
    goVersion: go1.19.3
    GOOS: darwin
    GOARCH: amd64
    numCPU: 12
    compiler: gc
    buildTime: "2023-01-06 16:13:54"
dependency:
    kclvmgoVersion: v0.4.4-beta.2
    kclPluginVersion: v0.4.1-alpha2
```

PR #229, opened 2023-02-01
bugfix: no such file or directory for compile

howieyuen:
close: #223

when compile KCL project, will default output to `ci-test/std.golden.yaml` when compile single KCL file, should default output to `stdout`


markliby (approved):
lgtm

Issue #204, opened 2022-12-27
[Feature] Request for KusionStack as a client-side solution for OAM


SparkYuan:
We can cooperate with KubeVela to make a more complete solution for our users

https://github.com/kubevela/kubevela/issues/5219

PR #238, opened 2023-02-08
feat: watch support kubevela Application from core.oam.dev/v1beta1

howieyuen:
#### What type of PR is this?

/kind feature

Support kubevela integration, after custom health check extension is ready, we can revert this PR.

**Note:**
kubevela Application API cannot imported by go module directly, because the replace statement will break go install.
I also tried the [kubevela-core-api](https://github.com/kubevela/kubevela-core-api), still get the same result.

related https://github.com/KusionStack/kusion/issues/204

```release-note
feat: watch support kubevela Application from core.oam.dev/v1beta1
```

```docs
None
```


third_party/kubevela/kubevela/apis/common/deepcopy_generated.go
```diff
@@ -0,0 +1,272 @@
+package common
+
+import (
+	v1 "k8s.io/api/core/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+
+	"kusionstack.io/kusion/third_party/kubevela/workflow/api/v1alpha1"
+)
+
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
SparkYuan, line 10: According to my understanding, we only need the kubevela `Application` definition to simplify our watch command. Do we need all these autogenerated DeepCopy codes and where will we use them?

howieyuen: Manifest(YAML) must be converted to Application(runtime.Object), so that health checker and printer know the field path.

SparkYuan: ok, I got it. DeepCopyObject() is an interface method of runtime.Object

howieyuen: yep

SparkYuan (approved):

PR #240, opened 2023-02-09
use reflect pkg to break the limit of for-range by dynamic select/case

### LINKED ISSUE #239: kusion watch can be optimized by reflect mechanism
## Enhancement

Currently, kusion will watch all resources from a single stack,  as well as the dependents resources, for example, watch `Deployment` as well as its `ReplicaSet` and `Pod`.

So the whole process basically can be handled by 2-level for-loop, 1st level is to range stack resources, 2nd is to range its depends resources. But for-range can be only done in order.

The `reflect` pkg offers dynamic `select` and `case` for slices type, enforce it will improve performance a lot.

Biggest change is for-loop is orderly read, reflect is dynamically read.

howieyuen:
#### What type of PR is this?

/kind feature

Fixes #239

```release-note
refactor: use reflect pkg to break the limit of for-range by dynamic select/case set
```

```docs
None
```

### DIFF
--- a/pkg/engine/operation/watch.go
+++ b/pkg/engine/operation/watch.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/gosuri/uilive"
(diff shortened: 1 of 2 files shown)


markliby (approved):
lgtm

Issue #241, opened 2023-02-09
kusion prevew flag `--ignore-fields` support more than map[string]interface{} type


howieyuen:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I wanna to ignore k8s Service of `spec.ports[*].targetPort`, but ports is a slice type, now kusion references `pkg/apis/meta/v1/unstructured`, its supports `map[string]interface{}` only.

```yaml
# ignore spec.ports.targetPort
apiVersion: v1
kind: Service
metadata:
  name: foo
  namespace: bar
spec:
  clusterIP: 172.16.128.40
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
  sessionAffinity: None
  type: ClusterIP
```

**Describe the feature you'd like:**

Support more then map type

**Describe alternatives you've considered:**

Extends unstructured pkg

**Teachability, Documentation, Adoption, Migration Strategy:**

Do research on how json path work

howieyuen:
> Besides, specify the resource kind or type can be even better.

ResourceNode is used to support both k8s and tf, now it has a field `Type` to do that, so any object is a manifest of k/v pairs, specify k8s kind may not proper.

PR #243, opened 2023-02-13
enhancement: ignore field support slice type

howieyuen:
#### What type of PR is this?

/kind feature

Fixes https://github.com/KusionStack/kusion/issues/241

Slice type passes tests, use recursive type assertion can do that.
`kind` can not be specified, because `ResourceNode` supports k8s and tf, `Type` field is used to differ them.

```release-note
enhancement: ignore field support slice type
```

```docs
None
```

### DIFF
--- a/pkg/engine/operation/graph/resource_node.go
+++ b/pkg/engine/operation/graph/resource_node.go
@@ -6,8 +6,6 @@ import (
 	"reflect"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-
 	"kusionstack.io/kusion/pkg/engine/models"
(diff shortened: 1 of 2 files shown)


healthjyk (commented):
1, add the option to specify the "type" may be better, more flexible;
2, if removeNestedField failed, why no error returned?
3, removeNestedField cannot handle nested field with type which is not map[string]interface{} or []interface{}, it's better to give clear indication of this func is only used for obj generated by json unmarshal, and show the process of json marshal and unmarshal in the test case.

healthjyk:
I saw the issue already, included the "type" is more useful for me.

howieyuen:
> I saw the issue already, included the "type" is more useful for me.

I'll try it again, see is there any better idea to make it.

healthjyk (approved):

PR #325, opened 2023-04-23
Handle incompatible k8s ingress api change since 1.26.0.

SparkYuan:
#### What type of PR is this?
/kind feature

1. add a new func to handle incompatible k8s ingress api change since 1.26.0.
2. update related go mods


liu-hm19 (approved):
LGTM

PR #332, opened 2023-04-26
Update: kusion deps: ignore all paths in --ignore option even if they appear in focus paths

amyXia1994:
#### What type of PR is this?

/kind update

Update: kusion deps: ignore all paths in --ignore option even if they appear in focus paths

The old behavior of `kusion deps` was: for entrance file list of each stack/project, if they contains the file that should be ignored(passed through `--ignore` option), then that ignored entrance file will not be regarded as that stack/project's upstream dependency.

That behavior was designed to avoid hotspot dependence like `https://github.com/KusionStack/konfig/blob/main/base/pkg/kusion_models/kube/render/render.k` which will involve almost all the frontend/backend files to a stack/project's upstream dependency.

Now there are scenarios that require ignoring paths even if they appear in focus paths.

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/deps/options.go
+++ b/pkg/cmd/deps/options.go
@@ -128,6 +128,17 @@ func (o *DepsOptions) Run() (err error) {
 		shouldIgnore := toSet(o.Ignore)
 		focusPaths := toSet(o.Focus)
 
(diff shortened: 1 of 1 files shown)


SparkYuan (approved):
lgtm

Issue #318, opened 2023-04-12
Standardize the indentation of Spec YAML


SparkYuan:
### What would you like to be added?

Standardize the indentation of Spec YAML

### Why is this needed?

Make YAML more standardized and easier to read

PR #341, opened 2023-05-11
chore: use yaml.v2 to keep indent to 2 spaces

howieyuen:
close: https://github.com/KusionStack/kusion/issues/318

### DIFF
--- a/pkg/engine/kcl.go
+++ b/pkg/engine/kcl.go
@@ -1,45 +1,39 @@
 package engine
 
 import (
-	"fmt"
-
-	yamlv2 "gopkg.in/yaml.v2"
-	yamlv3 "gopkg.in/yaml.v3"
+	"encoding/json"
 
 	kcl "kusionstack.io/kclvm-go"
 
 	"kusionstack.io/kusion/pkg/engine/models"
 	"kusionstack.io/kusion/pkg/log"
-	jsonUtil "kusionstack.io/kusion/pkg/util/json"
 )
 
 const MaxLogLength = 3751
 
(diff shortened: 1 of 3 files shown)


pkg/engine/kcl.go
```diff
@@ -1,45 +1,39 @@
...
 	"kusionstack.io/kusion/pkg/engine/models"
 	"kusionstack.io/kusion/pkg/log"
-	jsonUtil "kusionstack.io/kusion/pkg/util/json"
 )
 
 const MaxLogLength = 3751
 
-func ResourcesYAML2Spec(resourcesYAML []kcl.KCLResult) (*models.Spec, error) {
-	resources := []models.Resource{}
+func KCLResult2Spec(kclResults []kcl.KCLResult) (*models.Spec, error) {
```
SparkYuan, line 14: The variable name "kclResults" seems too generic and not a good way to name variables. It is suggested to use a more business-specific naming convention.

SparkYuan: Our discussion is focused on the variable name, rather than its type. "resourceYAML" is not a proper name, and "resource" may be a better fit in terms of semantics for the business.

howieyuen: `resource` is acceptable, but I still go with `KCLResult`, cause it is in `kcl.go`.

SparkYuan: What I mean is to use the semantics of Kusion instead of KCL, after all, it is invoked in the context of Kusion. Both ways are reasonable, you can choose the one you like

SparkYuan (approved):

Issue #276, opened 2023-03-08
Duplicated error message output when failed to destroy resources


liu-hm19:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

step 1 Try to destroy a nil resource and an oss bucket with contents in it. 
step 2 You can see the error message printed twice to the screen
### 2. What did you expect to see? (Required)
error message printed only once
### 3. What did you see instead (Required)
![image](https://user-images.githubusercontent.com/78610302/223649595-2faeb01c-16ad-48c4-9bfc-518633ca23b3.png)

### 4. What is your KusionStack components version? (Required)

releaseVersion: v0.7.2+87b7ea71
gitInfo:
    latestTag: v0.7.2
    commit: 87b7ea7162db777aa19a7f8be78bc2e41534f560
    treeState: clean
buildInfo:
    goVersion: go1.19.3
    GOOS: darwin
    GOARCH: arm64
    numCPU: 8
    compiler: gc
    buildTime: "2023-03-08 15:17:54"
dependency:
    kclvmgoVersion: v0.4.5
    kclPluginVersion: v0.4.4

PR #342, opened 2023-05-11
bugfix: dedup pint apply/destroy dag node executing error

howieyuen:
close: https://github.com/KusionStack/kusion/issues/276


liu-hm19 (approved):
LGTM

PR #343, opened 2023-05-11
support init with specified template 

### LINKED ISSUE #265: [Feature Request] Kusion provides API for integration tools
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now the extension is accomplishing its goals by executing the kusion binary for each time and then resolve the `stdout & stderr`, but since the stdout/stderr of kusion binary are not designed for tools integration, the output messages like this are disturbing:
image

the vscode-kusion extension requires kusion to provide APIs for tools to integrate kusion's capacities more elegantly and normatively. 

**Describe the feature you'd like:**

At least a few Go APIs of kusion are required:
- compile
- preview
- apply
- watch
- list projects
- list stacks
- list templates

### ListTemplates API

- API id：`init.ListTemplates`
- params:

| name | type | required | default value | description | example |
| ----- | ----- | -------- | ------------- | ----------- | --------- |
| online | boolean | optional | false | determines if kusion downloads templates from online repo | `true`, `false` |
| url | string | optional | `"https://github.com/KusionStack/kusion-templates"` | the online repo url, meaningful only when `online==true`|  |
| fsPath | string | optional |  | the fsPath to find the templates, meaningful only when `online == false` |`/path/to/templates` |

- returns:

| name | type | description | example |
| ----- | ----- | ----------- | -------- |
| templateList | Template[] | list of templates found | |
| err | Error | error code and message | `{"code":"invalidPath"， "message":"path not exists (or repo url not exists)"}` |

```go
type Template struct {
    name string
    description string
    content string // now the template could be the content of `kusion.yaml` file under the template directory: for example https://github.com/KusionStack/kusion-templates/blob/main/code-city/kusion.yaml
}

type Error struct {
    code  string
    message string
}
```

healthjyk:
#### What type of PR is this?
feature

support specifying template when kusion init. By using flag --template-name, you can specify the template you want to use, prompt will not exist.

Fixes #265 

None

```release-note
None
```

```docs
None
```

### DIFF
--- a/pkg/cmd/init/options.go
+++ b/pkg/cmd/init/options.go
@@ -18,12 +18,13 @@ import (
 const jsonOutput = "json"
 
 type InitOptions struct {
-	TemplateNameOrURL string
-	Online            bool
-	ProjectName       string
-	Force             bool
-	Yes               bool
-	CustomParamsJSON  string
+	TemplateRepoPath string
+	TemplateName     string
+	Online           bool
+	ProjectName      string
+	Force            bool
+	Yes              bool
+	CustomParamsJSON string
 }
 
 func NewInitOptions() *InitOptions {
@@ -32,31 +33,30 @@ func NewInitOptions() *InitOptions {
 
 func (o *InitOptions) Complete(args []string) error {
 	if o.Online { // use online templates, official link or user-specified link
-		o.TemplateNameOrURL = getURL(args)
+		o.TemplateRepoPath = templateRepoPathFromURL(args)
 	} else { // use offline templates, internal templates or user-specified local dir
-		path, err := getPath(args)
+		path, err := templateRepoPathFromLocal(args)
 		if err != nil {
 			return err
 		}
-		o.TemplateNameOrURL = path
+		o.TemplateRepoPath = path
 	}
 	return nil
 }
 
 func (o *InitOptions) Validate() error {
-	if o.Online {
-		return nil
-	}
-	// offline mode may need to generate templates
-	if err := validatePath(o.TemplateNameOrURL); err != nil {
-		return err
+	if !o.Online {
+		// offline mode may need to generate templates
+		if err := validateLocalTemplateRepoPath(o.TemplateRepoPath); err != nil {
+			return err
+		}
 	}
 	return nil
 }
 
 func (o *InitOptions) Run() error {
 	// Retrieve the template repo.
-	repo, err := retrieveTemplateRepo(o.TemplateNameOrURL, o.Online)
+	repo, err := retrieveTemplateRepo(o.TemplateRepoPath, o.Online)
 	if err != nil {
 		return err
 	}
@@ -70,8 +70,23 @@ func (o *InitOptions) Run() error {
 
 	// Choose template
 	var template scaffold.Template
-	if template, err = chooseTemplate(templates); err != nil {
-		return err
+	if o.TemplateName == "" {
+		// if template name is not specified, choose template by prompt
+		if template, err = chooseTemplate(templates); err != nil {
+			return err
+		}
+	} else {
+		// if template name is specified, find template from repo data
+		var templateExist bool
+		for _, t := range templates {
+			t.Name = o.TemplateName
+			templateExist = true
+			template = t
+			break
+		}
+		if !templateExist {
+			return fmt.Errorf("template %s does not exist", o.TemplateName)
+		}
 	}
 
 	// Parse customParams if not empty
@@ -182,10 +197,9 @@ func (o *InitOptions) Run() error {
 }
 
 type TemplatesOptions struct {
-	Online bool
-	URL    string
-	Path   string
-	Output string
+	TemplateRepoPath string
+	Online           bool
+	Output           string
 }
 
 func NewTemplatesOptions() *TemplatesOptions {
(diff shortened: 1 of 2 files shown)


pkg/cmd/init/options.go
```diff
@@ -18,12 +18,13 @@ import (
 const jsonOutput = "json"
 
 type InitOptions struct {
-	TemplateNameOrURL string
-	Online            bool
-	ProjectName       string
-	Force             bool
-	Yes               bool
-	CustomParamsJSON  string
+	TemplateRepoPath string
```
howieyuen, line 21: `TemplateNameOrURL` is miss understanding, that's my fault, this field contains online/offline mode, online is url, offline is local dir.
```suggestion
	TemplatePathOrURL string
````

File: pkg/cmd/init/options.go [hunk 2]
```diff
@@ -247,18 +256,18 @@ func (o *TemplatesOptions) Run() error {
 	return nil
 }
 
-// getURL parses url from args, called when --online is true.
-func getURL(args []string) string {
+// templateRepoPathFromURL parses url from args, called when --online is true.
+func templateRepoPathFromURL(args []string) string {
```
howieyuen, line 260: no need to change, `getURL` works for online mode only

pkg/cmd/init/options.go
```diff
@@ -289,8 +298,8 @@ func validatePath(path string) error {
 }
 
 // retrieveTemplateRepo gets template repos from online or local, with specified url or path.
-func retrieveTemplateRepo(templateName string, online bool) (scaffold.TemplateRepository, error) {
-	return scaffold.RetrieveTemplates(templateName, online)
+func retrieveTemplateRepo(templateRepoPath string, online bool) (scaffold.TemplateRepository, error) {
```
howieyuen, line 301:
```suggestion
func retrieveTemplateRepo(templatePathOrURL string, online bool) (scaffold.TemplateRepository, error) {
```

pkg/cmd/init/options.go
```diff
@@ -247,18 +256,18 @@ func (o *TemplatesOptions) Run() error {
...
 		return args[0]
 	}
 	return "" // use official link
 }
 
-// getPath parses path from args, if not specified, use default InternalTemplateDir,
+// templateRepoPathFromLocal parses path from args, if not specified, use default InternalTemplateDir,
 // called when --online is false.
-func getPath(args []string) (string, error) {
+func templateRepoPathFromLocal(args []string) (string, error) {
```
howieyuen, line 270: if miss understanding, `getLocalPath` can be better

pkg/cmd/init/options.go
```diff
@@ -272,8 +281,8 @@ func getPath(args []string) (string, error) {
 	}
 }
 
-// validatePath checks the path is valid or not.
-func validatePath(path string) error {
+// validateLocalTemplateRepoPath checks the path is valid or not.
+func validateLocalTemplateRepoPath(path string) error {
```
howieyuen, line 285: `validatePath` works offline mode only, if miss understanding, `validateLocalPath` can be better

pkg/cmd/init/options.go
```diff
@@ -32,31 +33,30 @@ func NewInitOptions() *InitOptions {
...
 }
 
 func (o *InitOptions) Validate() error {
-	if o.Online {
-		return nil
-	}
-	// offline mode may need to generate templates
-	if err := validatePath(o.TemplateNameOrURL); err != nil {
-		return err
+	if !o.Online {
```
howieyuen, line 48: nesting `if` statements isn't a cool code style

howieyuen (changes requested):

pkg/cmd/init/options.go
healthjyk: that's right, but TemplatePathOrURL is not elegant either. TemplateRepoPath is used to refer to the path which can get template repo, can be url or local path.

pkg/cmd/init/options.go
healthjyk: templateRepoFromOnline is better, thanks :)

pkg/cmd/init/options.go
healthjyk: actually, I think it's okay, maybe you can give me an example :)

pkg/cmd/init/options.go
howieyuen: `TemplateRepoPath` is a path, when does `path` means URL?

pkg/cmd/init/options.go
howieyuen: original style is an example, it's called "guard clauses"

pkg/cmd/init/options.go
howieyuen: it refers a local path, which should contain templates, but only validate it's a dir, not a file, or even not exists.
this function doesn't check whether templates are valid or not.

pkg/cmd/init/options.go
healthjyk: haha, the original code is also provided by me. I changed it because maybe there will be validation when online is true. Your suggestion is better for now, I'll change it.

pkg/cmd/init/options.go
healthjyk: haha, the original code is also provided by me. I changed it because maybe there will be validation when online is true. Your suggestion is better for now, I'll change it.

pkg/cmd/init/options.go
healthjyk: That’s what I want to express, is it confusing?

amyXia1994 (dismissed):
lgtm

pkg/cmd/init/options.go
healthjyk: yeah, you are right, so the suffix is "Path". "LocalTemplateRepoPath" asks the path refer to a local template repo, although the func only checks it's a dir or not for now, more validation for local template repo can be added in the future.

howieyuen (approved):
/lgtm

PR #346, opened 2023-05-15
support kusion preview --ouput json

### LINKED ISSUE #265: [Feature Request] Kusion provides API for integration tools

healthjyk:
#### What type of PR is this?
feature

when using cmd kusion preview, there is no way to get the pure preview result, cause there is prompt. Support --output json to print the preview result without any unnecessary prompt when succeeded.

Fixes #265 

```release-note
the output is json marshal of struct pkg/engine/operation/models.Changes
```

```docs
None
```

### DIFF
--- a/pkg/cmd/preview/options.go
+++ b/pkg/cmd/preview/options.go
@@ -1,6 +1,8 @@
 package preview
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 
@@ -21,6 +23,8 @@ import (
 	"kusionstack.io/kusion/pkg/util/pretty"
 )
 
+const jsonOutput = "json"
+
 type PreviewOptions struct {
 	compilecmd.CompileOptions
 	PreviewFlags
@@ -32,6 +36,7 @@ type PreviewFlags struct {
 	Detail       bool
 	All          bool
 	NoStyle      bool
+	Output       string
 	IgnoreFields []string
 }
 
@@ -46,7 +51,13 @@ func (o *PreviewOptions) Complete(args []string) {
 }
 
 func (o *PreviewOptions) Validate() error {
-	return o.CompileOptions.Validate()
+	if err := o.CompileOptions.Validate(); err != nil {
+		return err
+	}
+	if o.Output != "" && o.Output != jsonOutput {
+		return errors.New("invalid output type, supported types: json")
+	}
+	return nil
 }
 
 func (o *PreviewOptions) Run() error {
@@ -54,6 +65,10 @@ func (o *PreviewOptions) Run() error {
 	if o.NoStyle {
 		pterm.DisableStyling()
 		pterm.EnableColor()
+	}
+	if o.Output == jsonOutput {
+		pterm.DisableStyling()
+		pterm.DisableColor()
 	}
 
 	// Parse project and stack of work directory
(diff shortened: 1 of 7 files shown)


pkg/cmd/preview/options.go
```diff
@@ -72,14 +87,20 @@ func (o *PreviewOptions) Run() error {
...
 	}, project, stack)
 	if err != nil {
 		return err
 	}
 
 	// return immediately if no resource found in stack
 	if sp == nil || len(sp.Resources) == 0 {
-		fmt.Println(pretty.GreenBold("\nNo resource found in this stack."))
+		if o.Output == jsonOutput {
+			emptyChanges, _ := json.Marshal(&opsmodels.Changes{})
```
SparkYuan, line 99: We can maintain the original output message to provide users with a friendly hint

amyXia1994 (approved):
lgtm

pkg/cmd/preview/options.go
SparkYuan: Maybe you can refer to the approach of Kubectl for non-existent resources. I think their approach is more user-friendly.

image

healthjyk: When errored, test-format output is shown; but if succeeded, the specified format is get output.

healthjyk: The left feature upgrade is that even if there is an empty spec, diff should still be done, cause it means delete.

healthjyk:
> lgtm

@amyXia1994 if there is no resource in the planned spec, there will be no output under --output json, you may need a judgement before json unmarshal.

healthjyk:
> lgtm



> lgtm

@amyXia1994 If there is no resource in the planned spec, there will be an empty output when --output json, you may need a judgement before json unmarshal. For the case without --output json, there is a hint, but still with successful preview.

SparkYuan (approved):
lgtm

PR #363, opened 2023-06-27
chore(deps): delete the gookit/goutil dependency

SparkYuan:
#### What type of PR is this?

/kind cleanup

delete the gookit/goutil dependency

We don't need it anymore and it contains security issues.


liu-hm19 (approved):
lgtm

PR #367, opened 2023-07-07
feat: refine no-style and remove no-prompt command option

### LINKED ISSUE #366: Optimize the display of plain text output of kusion cmd for better integration with other systems

Add the "no-style" option to kusion compile to facilitate integration with other systems, while optimizing the display of pure text error messages.

### Why is this needed?

Owing to the difficulty of reading output info with color (say like in ANSI format) in log message, it is necessary to optimize the display of plain text output of kusion command line (add "no-style" option to kusion compile for example), which would facilitate the integration with other systems. Besides, the plain text error messages should also be refined. 
image

liu-hm19:
#### What type of PR is this?

/kind feature

This PR refines "no-style" command line option. 

Fixes #366 

This PR removes "no-prompt" command line option. 

```release-note
Add "no-style" option for kusion compile for better third-party integration while remove the "no-prompt" option. 
```

```docs

```

### DIFF
--- a/pkg/cmd/spec/generator.go
+++ b/pkg/cmd/spec/generator.go
@@ -3,7 +3,9 @@ package spec
 import (
 	"fmt"
 
+	"github.com/acarl005/stripansi"
 	"github.com/pterm/pterm"
+	"sigs.k8s.io/kustomize/kyaml/errors"
 
 	"kusionstack.io/kusion/pkg/engine/models"
 	"kusionstack.io/kusion/pkg/generator"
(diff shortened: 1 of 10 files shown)


liu-hm19:
Here are some cmd examples after the optimization: 
image
image

pkg/cmd/spec/generator.go
```diff
@@ -3,7 +3,9 @@ package spec
 import (
 	"fmt"
 
+	"github.com/acarl005/stripansi"
 	"github.com/pterm/pterm"
+	"sigs.k8s.io/kustomize/kyaml/errors"
```
SparkYuan, line 8: why do we use a kustomize error instead of the native go errors

pkg/cmd/spec/generator.go
```diff
@@ -42,16 +43,21 @@ func GenerateSpecWithSpinner(o *generator.Options, project *projectstack.Project
 
 	spec, err := g.GenerateSpec(o, stack)
 	if err != nil {
-		if !o.NoPrompt && sp != nil {
+		if !o.NoStyle && sp != nil {
 			sp.Fail()
 		}
+
+		if o.NoStyle {
+			return nil, errors.Wrap(stripansi.Strip(err.Error()))
```
SparkYuan, line 51: please add a todo comment here to indicate that we will replace this with KCL no-style flags when it is supported

pkg/cmd/spec/generator.go
liu-hm19: I've changed to the native go errors package.

pkg/cmd/spec/generator.go
liu-hm19: OK, I've  added the comment : )

SparkYuan (approved):
lgtm

PR #428, opened 2023-08-02
refactor: use ordered map for component and container

elliotxx:
#### What type of PR is this?
/kind refactor

Use ordered map for component and container. Otherwise, each compilation result will be different.


adohe (approved):
/lgtm

PR #453, opened 2023-08-14
feat: modify the generators to adapt to the new model's adjustments

elliotxx:
#### What type of PR is this?
/kind feature

Modify the go model of AppConfiguration, generators to adapt to the new model's adjustments.

New model:

https://github.com/KusionStack/catalog/blob/main/models/schema/v1/app_configuration.k


pkg/generator/appconfiguration/app_configurations_generator.go
```diff
@@ -21,11 +21,12 @@ func (acg *Generator) GenerateSpec(
 		Resources: []models.Resource{},
 	}
 
-	g, err := generators.NewAppConfigurationGenerator(project.Name, acg.AppConfiguration)
-	if err != nil {
-		return nil, err
-	}
-	if err = g.Generate(spec); err != nil {
+	gfs := []generators.NewGeneratorFunc{}
+	generators.ForeachOrderedApps(acg.Apps, func(appName string, app appmodel.AppConfiguration) error {
```
SparkYuan, line 25: It's a little over-designed that we add a function into the params. Considering that we only invoked this function once, we can inline the `func(appName string, app appmodel.AppConfiguration) error` in the implication of `ForeachOrderedApps`

SparkYuan (approved):
overall, it looks good to me

pkg/generator/appconfiguration/app_configurations_generator.go
elliotxx: Next PR, I will optimize this method using `Generics`.

PR #457, opened 2023-08-16
fix: missing labels and annotations injection logic

elliotxx:
#### What type of PR is this?
/kind bug

- Fix missing labels and annotations injection logic
- Fix some typo


adohe (approved):
/lgtm

PR #484, opened 2023-08-22
test: complete appconfiguartion model unit tests

elliotxx:
#### What type of PR is this?
/kind test

* Complete appconfiguartion model unit tests
* Fix wrong acceptor type of MarshalJSON and MarshalYAML

### DIFF
--- a/pkg/models/appconfiguration/workload/workload.go
+++ b/pkg/models/appconfiguration/workload/workload.go
@@ -22,7 +22,7 @@ type Workload struct {
 	*Job     `yaml:",inline" json:",inline"`
 }
 
-func (w *Workload) MarshalJSON() ([]byte, error) {
+func (w Workload) MarshalJSON() ([]byte, error) {
 	switch w.Header.Type {
 	case TypeService:
 		return json.Marshal(struct {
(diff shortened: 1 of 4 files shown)


pkg/models/appconfiguration/workload/workload.go
```diff
@@ -22,7 +22,7 @@ type Workload struct {
 	*Job     `yaml:",inline" json:",inline"`
 }
 
-func (w *Workload) MarshalJSON() ([]byte, error) {
```
SparkYuan, line 25: why do you replace the reference with instance? All structs containing `Workload` are used as a reference.

healthjyk: Test comment

elliotxx: Using `(w *Workload)` cannot serialize `Workload{}` instance as expected, only serialize `&Workload{}`.
However, using `(w Workload)` can serialize `Workload{}` and `&Workload{}` instance as expected.
You can refer to the `MarshalJSON` and `UnmarshalJSON` implementation of `Time` in the standard library:
https://sourcegraph.com/github.com/golang/go/-/blob/src/time/time.go?L1360:15

SparkYuan: ok, i got it

SparkYuan (approved):
lgtm

PR #491, opened 2023-08-24
feat: add lifecycle type and encoding/decoding support

adohe:
#### What type of PR is this?
/kind feature

update Container model add lifecycle types and encoding/decoding support.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/models/appconfiguration/workload/container/container.go
+++ b/pkg/models/appconfiguration/workload/container/container.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 )
 
+// Container describes how the Application's tasks are expected to be run.
 type Container struct {
 	// Image to run for this container
 	Image string `yaml:"image" json:"image"`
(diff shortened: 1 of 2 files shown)


pkg/models/appconfiguration/workload/container/container.go
```diff
@@ -44,16 +68,11 @@ type Probe struct {
...
-	// Type of action to be taken.
-	Type string `yaml:"_type" json:"_type"`
-}
-
 // ProbeHandler defines a specific action that should be taken in a probe.
 // One and only one of the fields must be specified.
 type ProbeHandler struct {
 	// Type of action to be taken.
-	ProbeType `yaml:"_type" json:"_type"`
+	ActionType `yaml:"_type" json:"_type"`
```
SparkYuan, line 75: Can you explain the reason behind renaming `ProbeType` to `ActionType`? According to my understanding, `_type` is the schema name of kusion models, maybe you can just name this field as `Type`.

adohe: To make YAML decoder happy, we have to add a type wrapper struct, and for Golang anonymous field requirement, we cannot name this wrapper struct as `Type`

SparkYuan (approved):
lgtm

PR #498, opened 2023-08-28
fix: error messages are ignored when executing resource nodes

SparkYuan:
#### What type of PR is this?

/kind bug

fix: error messages are ignored when executing resource nodes


panshuai-ps (approved):
lgtm

PR #501, opened 2023-08-30
fix: port in probe url now gets parsed properly

ffforest:
#### What type of PR is this?

/kind bug

The URL in Probe model gets parsed and put into the corev1.Probe object.
The port gets parsed by `intstr.FromString(u.Port())` which ends up being a string in the rendered yaml:
image
This is rejected by Kubernetes because for a string port (as the port name) it expects at least one letter in it:
image
[`intstr.Parse()`](https://pkg.go.dev/k8s.io/apimachinery@v0.27.2/pkg/util/intstr#Parse) will attempt to convert the string to an int first which is what we want.

Fixes #

```release-note

```

```docs

```


adohe (approved):
/lgtm

PR #510, opened 2023-09-06
fix: add a kusion-control label in svc labels

SparkYuan:
#### What type of PR is this?
/kind bug

add a kusion-control label in svc labels

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
+++ b/pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
@@ -19,9 +19,10 @@ const (
 	suffixPublic   = "public"
 	suffixPrivate  = "private"
 
-	// aliyun SLB annotations, ref: https://help.aliyun.com/zh/ack/ack-managed-and-ack-dedicated/user-guide/add-annotations-to-the-yaml-file-of-a-service-to-configure-clb-instances?spm=a2c4g.11186623.0.0.59e26219ESUbqe
+	// aliyun SLB annotations, ref: https://help.aliyun.com/zh/ack/ack-managed-and-ack-dedicated/user-guide/add-annotations-to-the-yaml-file-of-a-service-to-configure-clb-instances
 	aliyunLBSpec     = "service.beta.kubernetes.io/alibaba-cloud-loadbalancer-spec"
 	aliyunSLBS1Small = "slb.s1.small"
+	kusionControl    = "kusionstack.io/control"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


liu-hm19:
lgtm

liu-hm19 (approved):

pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
```diff
@@ -163,16 +163,13 @@ func (g *portsGenerator) generateK8sSvc(public bool, ports []network.Port) *v1.S
 
 		// only support Aliyun SLB for now, and set SLB spec by default.
 		svc.Annotations[aliyunLBSpec] = aliyunSLBS1Small
+		svc.Labels[kusionControl] = "true"
```
healthjyk, line 166: why you need to add this label? K8s Service will Inherit the label that Collaset/Deployment have, does they don't have the label of kusionControl?

SparkYuan: This label is used to cooperate with the KAFE svc controller to implement gracefully rollout. The KAFE svc controller will only process svcs with this label and it is given in the svc sample before.

pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
```diff
@@ -163,16 +163,13 @@ func (g *portsGenerator) generateK8sSvc(public bool, ports []network.Port) *v1.S
...
 	}
 
 	return svc
 }
 
 func validatePorts(ports []network.Port) error {
-	if len(ports) == 0 {
-		return ErrEmptyPorts
-	}
-
```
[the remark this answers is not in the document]
SparkYuan: If the ports are nil, the logic below will return immediately. I don't understand why this generator will fail.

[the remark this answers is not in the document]
healthjyk: 1, I don't know what is "the logic below" you refer to. Besides, what will get returned? an error, or an empty result?
2, If not k8s svc will get generated, why you new the ports generator? It will generate nothing. Add a judgement in “service generator" is much more concise and correct.

pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
healthjyk: I know this. But I want to know does this label is not attached to the KAFE workload, Collaset? Does it only need to attach to the service?
This label works only for KAFE system, should not get added under other situation.

pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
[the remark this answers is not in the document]
SparkYuan: 1. The logic below this line. That means the logic in this function. This function will return nil if the ports are empty
2. Yes, deliberately return before the validation function is more readable

pkg/generator/appconfiguration/generator/workload/network/ports_generator.go
SparkYuan: It only works with the KAFE workload. We will move this label and all Aliyun-related labels to platform.k in the next version

PR #522, opened 2023-09-11
feat: use map slice for container env

### LINKED ISSUE #503: Toolchain: Workload generator should manage the order of the environment variables properly

Currently workload(service and job) generator uses a map (which is unordered) to store and iterate over all the environment variables, including:
1. User provided env vars
2. Platform environment variables that kusion injects automatically (such as database credentials, etc)

The environment variables need to be in the correct order for this to work properly.

### Why is this needed?

The expected order is:
- All the kusion-managed env vars should be placed first
- The user-provided env vars comes after because they might be referencing kusion-managed env vars (e.g. WORDPRESS_DB_HOST=$(KUSION_DB_HOST)
- The user-provided env vars should appear in the kubernetes yaml in the same order as they appear in the KCL configuration, because user might also want to leverage the A=$(B) expression as supported in Kubernetes: https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/

liu-hm19:
#### What type of PR is this?

/kind feature
This PR uses MapSlice in yaml.v2 package for container env so that the order of the environment variables can stay consistent. 

Fixes #503 

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/spec/generator.go
+++ b/pkg/cmd/spec/generator.go
@@ -120,6 +120,9 @@ func GenerateSpecFromFile(filePath string) (*models.Spec, error) {
 		return nil, err
 	}
 
+	// TODO: here we use decoder in yaml.v3 to parse resources because it converts
+	// map into map[string]interface{} by default which is inconsistent with yaml.v2.
+	// The use of yaml.v2 and yaml.v3 should be unified in the future.
 	decoder := yamlv3.NewDecoder(bytes.NewBuffer(b))
 	decoder.KnownFields(true)
 	var resources models.Resources
--- a/pkg/cmd/spec/generator.go
+++ b/pkg/cmd/spec/generator.go
@@ -103,6 +103,9 @@ func buildAppConfigs(o *generator.Options, stack *projectstack.Stack) (map[strin
 
 	log.Debugf("unmarshal %s to app configs", out)
 	appConfigs := map[string]appmodel.AppConfiguration{}
+
+	// Note: we use the type of MapSlice in yaml.v2 to maintain the order of container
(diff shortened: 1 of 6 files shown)


pkg/cmd/spec/generator.go
```diff
@@ -9,7 +9,8 @@ import (
 
 	"github.com/acarl005/stripansi"
 	"github.com/pterm/pterm"
-	"gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v2"
+	yamlv3 "gopkg.in/yaml.v3"
```
SparkYuan, line 13: Can you explain the differences between the yamlv2 and yamlv3 modules that you have imported?

liu-hm19: Sure, yaml.v2 has the type of `MapSlice` which we are using to maintain the order of container's environment variables. However, yaml.v3 doesn't has this type, and thus we should import both of the packages here.

liu-hm19: Sure, yaml.v2 has the type of `MapSlice` which we are using to maintain the order of container's environment variables. However, yaml.v3 doesn't has this type, and thus we should import both of the packages here.

SparkYuan: If yamlv2 is necessary, please add some comments about it. Can we change yamlv3 to yamlv2 in this file?

liu-hm19: OK, I've added some comments. It's sad that we can't just replace yaml.v3 with yaml.v2 in this file because the method `KnownFields` called in `GenerateSpecFromFile` is undefined in yaml.v2.

PR #532, opened 2023-09-11
fix: wrong e2e ut for kusion deps and apply

elliotxx:
#### What type of PR is this?
/kind bug

Fix wrong e2e ut for kusion deps and apply


adohe (approved):
/lgtm

PR #548, opened 2023-09-14
fix: delete external templates when init --online failed

### LINKED ISSUE #368: kusion init --online failed
### What happened?

When I use kusion init --online, an error “cloning templates repos: reference not found". Is it the problem of directory?
image

Fix it

run kusion init --online

_No response_

### Kusion version

<details>

```console
$ kusion version
# paste output here
```

</details>
v0.8.1

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

### Install tools

<details>

</details>

SparkYuan:
#### What type of PR is this?
/kind bug

always delete external templates in init with the online flag

Fixes #368

```release-note

```

```docs

```

### DIFF
--- a/pkg/scaffold/templates.go
+++ b/pkg/scaffold/templates.go
@@ -2,6 +2,7 @@ package scaffold
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -10,9 +11,7 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/gitutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 	"github.com/spf13/afero"
@@ -26,7 +25,7 @@ import (
 // These are variables instead of constants in order that they can be set using the `-X`
 // `ldflag` at build time, if necessary.
 var (
-	// The Git URL for Kusion program templates
+	// KusionTemplateGitRepository is the Git URL for Kusion program templates
 	KusionTemplateGitRepository = "https://github.com/KusionStack/kusion-templates"
 	// The branch name for the template repository
 	kusionTemplateBranch = "main"
@@ -104,7 +103,7 @@ func LoadTemplate(path string) (Template, error) {
 		return Template{}, err
 	}
 	if !info.IsDir() {
-		return Template{}, errors.Errorf("%s is not a directory", path)
+		return Template{}, fmt.Errorf("%s is not a directory", path)
 	}
 
 	proj, err := LoadProjectTemplate(filepath.Join(path, KusionYaml))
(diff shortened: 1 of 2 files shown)


pkg/scaffold/templates.go
```diff
@@ -192,23 +191,23 @@ func retrieveFileTemplates(path string) (TemplateRepository, error) {
...
 	templateDir, err := GetTemplateDir(ExternalTemplateDir)
 	if err != nil {
 		return TemplateRepository{}, err
 	}
 
+	// delete old external templates
+	if err = os.RemoveAll(templateDir); err != nil {
+		return TemplateRepository{}, err
+	}
+
```
healthjyk, line 208: It seems that removing repos if git clone failed is much better, add a `defer` after line 218.
Besides, why using pulumi to git clone, it's such a basic function. Use a third-party package may cause unnecessary dependency problems.

pkg/scaffold/templates.go
```diff
@@ -351,15 +314,15 @@ var (
...
 	}
 
 	if len(s) > 100 {
-		return errors.New("A project name must be 100 characters or less")
+		return errors.New("a project name must be 100 characters or less")
 	}
 
 	if !validProjectNameRegexp.MatchString(s) {
-		return errors.New("A project name may only contain alphanumeric, hyphens, underscores, and periods")
+		return errors.New("a project name may only contain alphanumeric, hyphens, underscores, and periods")
```
healthjyk, line 325: not "may", but "must".

pkg/scaffold/templates.go
```diff
@@ -351,15 +314,15 @@ var (
 // for display to an end user.
 func ValidateProjectName(s string) error {
 	if s == "" {
-		return errors.New("A project name may not be empty")
+		return errors.New("a project name may not be empty")
```
healthjyk, line 317: not "may", but "must".
delete "a" is better, or "a project's name", the same for the below.

pkg/scaffold/templates.go
```diff
@@ -351,15 +314,15 @@ var (
...
 	}
 
 	if len(s) > 100 {
-		return errors.New("A project name must be 100 characters or less")
+		return errors.New("a project name must be 100 characters or less")
```
healthjyk, line 321: "must be equal or less than 100 characters“ is better

pkg/scaffold/templates.go
SparkYuan: Are you intending to add `defer` with `recover` in this function? It will catch all errors and we have to distinguish which one is a git error.I moved the logic to #L215. The template directory will only be removed if the git clone fails.

pkg/scaffold/templates.go
```diff
@@ -192,33 +191,31 @@ func retrieveFileTemplates(path string) (TemplateRepository, error) {
...
 
 	if online {
 		// Clone or update the kusion/templates repo.
 		repo := KusionTemplateGitRepository
 		branch := plumbing.NewBranchReferenceName(kusionTemplateBranch)
-		err := gitutil.GitCloneOrPull(repo, branch, templateDir, false /*shallow*/)
+		err = gitutil.GitCloneOrPull(repo, branch, templateDir, false /*shallow*/)
 		if err != nil {
-			return TemplateRepository{}, fmt.Errorf("cloning templates repo: %w", err)
+			if err2 := os.RemoveAll(templateDir); err2 != nil {
```
healthjyk, line 215: maybe removeErr is better.

healthjyk:
please upgrade the title

healthjyk (approved):

PR #594, opened 2023-11-09
docs: add kusion environment management proposal

### LINKED ISSUE #589: Feat: Kusion environment management command

Add command `kusion env` to support env init and management

### Why is this needed?

We have added environment management into the platform engineer workflow, so we should add a kusion command to help platform developers to complement their work.

healthjyk:
#### What type of PR is this?
/kind documentation

add doc "kusion environment management proposal".

Fixes #589


docs/design/enviroment_management.md
```diff
@@ -0,0 +1,156 @@
...
+
+The upgrade explicitly the role of platform engineers and the concepts of **Module** and **Environment**. The Module is a building block enabling the delivery of a set of resources from **Schema Configuration** and **Environment Configuration**. While the Environment is a collection of application-independent configurations, and an environment configuration is composed of a set of module inputs, which forms the entire module inputs combined with schema configuration.
+
+<br />![module](module.jpg)
```
SparkYuan, line 9: This diagram is confusing, what is the relationship between schema, input, output, and en var? Additionally, based on my understanding, provider is not a part of the Kusion module.

docs/design/enviroment_management.md
```diff
@@ -0,0 +1,156 @@
...
+# Module input, with format standard：
+# # module_<ModuleName>:
+# #   default: # default configurations, applied to all projects
+# #     <ModuleInput_Field1>: <value1>
+# #     <ModuleInput_Field2>: <value2>
+# #     ...
+# #   <overrideGroupName>: # override configurations, applied to the projects assigned in projectSelector
+# #     <ModuleInput_Field1>: <value1_Override>
+# #     ... 
+module_Database:
```
SparkYuan, line 38: Why should the module block name have a `module` prefix?  It is also awkward that the module block's name is in the snake format and the `projectSelector` is in the camel format.

docs/design/enviroment_management.md
```diff
@@ -0,0 +1,156 @@
...
+  baz:
+    instanceClass: db.t3.large
+    projectSelector:
+    - baz  
+```
+**A1**: The format shown above can work, but there exists some drawbacks:
+
+- projectSelector becomes a keyword, it asks for the module input filed name must not be projectSelector;
+- projectSelector is at the same layer of the module input, which brings possible misunderstanding;
+- Not obey 3NF of [Database normalization](https://en.wikipedia.org/wiki/Database_normalization), there are non-key columns (module input fields) depends on other non-key columns (projectSelector), which is not a good design for the possible database storage backend in the future.
```
SparkYuan, line 144: I didn't understand the relationship between the `projectSelector` and 3NF, can you provide some examples?

docs/design/environment/enviroment_management.md
```diff
@@ -0,0 +1,153 @@
...
+
+At the product level, Kusion does not support multiple clouds or multiple regions in an environment. Although users can define an artful module which builds a schema into multiple resources in multiple clouds or regions. Kusion cannot forbid such modules, but doesn't recommend it and will not provide supportive technical means of multiple clouds or region in an environment. If a platform team needs to manage multiple clouds or regions, just create multiple environments. 
+
+Environment config are kept by yaml file when stored locally, and only includes a set of module inputs for now. But it may include other data, such as backend, which is not appropriate to set in the configuration file. There are two reasons: 1. backend info is a global config rather than environment config; 2. backend info includes sensitive data, such as database password.
```
SparkYuan, line 33: Based on my experience, backend configurations typically vary across different environments. For example, the OSS path for prod is not the same as dev. Hence, it is reasonable to set up backend configs in the environment config.

docs/design/environment/enviroment_management.md
```diff
@@ -0,0 +1,153 @@
...
+# #     ...
+module:
+  database:
+    default:
+      provider: aws
+      size: 20
+      instanceClass: db.t3.micro
+      securityIPs:
+      - 10.0.0.0/18
+    smallClass:
```
SparkYuan, line 66: Besides the override function, can this patcher add a new k-v pair into the default block?

docs/design/environment/enviroment_management.md
```diff
@@ -0,0 +1,153 @@
...
+### Non-Goals
+
+1. Only support local file based environment management, the support of other storage backends is not the goal of version 0.10.0;
+2. Only consider using Kusion in CLI format, the usage of service layer is not considered.
+
+## Proposal
+
+### Definition
+
+Environment is a logical concept representing a target that a stack will be deployed to, which contains a set of configurations, kubeconfig and provider authentication information that can be reused by multiple stacks. We recommend environments grouped by SDLC phases or cloud vendors. For example, it could be named as dev, staging, and prod or cloud vendors such as AWS, Azure, and Aliyun.
```
ffforest, line 29: > it could be named as dev, staging, and prod or cloud vendors such as AWS, Azure, and Aliyun

or could be both

docs/design/environment/enviroment_management.md
```diff
@@ -0,0 +1,153 @@
...
+
+At the product level, Kusion does not support multiple clouds or multiple regions in an environment. Although users can define an artful module which builds a schema into multiple resources in multiple clouds or regions. Kusion cannot forbid such modules, but doesn't recommend it and will not provide supportive technical means of multiple clouds or region in an environment. If a platform team needs to manage multiple clouds or regions, just create multiple environments.
```
ffforest, line 31: > At the product level, Kusion does not support multiple clouds or multiple regions in an environment. Although users can define an artful module which builds a schema into multiple resources in multiple clouds or regions. Kusion cannot forbid such modules, but doesn't recommend it and will not provide supportive technical means of multiple clouds or region in an environment.

Do we need to mention this explicitly? This is more or less part of our thought process.

docs/design/environment/enviroment_management.md
```diff
@@ -0,0 +1,153 @@
...
+List all the initialized environments.
+
+#### kusion env show \<name>
+
+Show the configurations of specified environment.
+
+There are the user cases:
+
+- **Individual developer**: The user creates environments as needed (e.g. dev, pre and prod), and writes AppConfiguration codes in corresponding project and stacks locally. When applying a specified stack, set kubeconfig and provider authentication info, and corresponding environment configurations are used automatically to build intent.
+- **Collaboration Scenario**: An application developer gets the names of environments in use, then write project and stack AppConfiguration codes locally, with stack name the same as environment name. In CI/CD pipeline, create environments according to the changed stacks, set kubeconfig and provider authentication info, then apply a specified stack. In this scenario, the full environment configurations should be stored by platform team for now. After the support of centralized storage backend, the storage work of environment configuration can be handled by Kusion.
```
ffforest, line 116: > In CI/CD pipeline, create environments according to the changed stacks, set kubeconfig and provider authentication info, then apply a specified stack.

Just thinking, if the pipeline code needs to pull down the file manually and perform `kusion env create` every time the pipeline runs, conceptually it kind of loses the point of an "environment" - which is an abstraction layer for the configs/secrets that are associated with it (Instead, the pipeline code directly manages the configs and secrets that are needed for Kusion modules).

docs/design/environment/enviroment_management.md
healthjyk: You're tight, an environment could be like AWS_dev

docs/design/environment/enviroment_management.md
healthjyk: In our website towards users, I think these explanation could not be mentioned.

docs/design/environment/enviroment_management.md
healthjyk: That's because we don't provide centralized storage for now. But for the command, we give a systematized design, and provide a good and simple environment management experience.

docs/design/environment/enviroment_management.md
```diff
@@ -0,0 +1,153 @@
...
+
+The upgrade defines the role of platform engineers and the concepts of **Module** and **Environment**. The Module is a building block enabling the delivery of a set of resources from **Schema Configuration** and **Environment Configuration**. While the Environment is a collection of application-independent configurations, and an environment configuration is composed of a set of module inputs, which forms the entire module inputs combined with schema configuration.
+
+<br />![workflow](workflow.png)
```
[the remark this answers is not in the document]
healthjyk: It seems we could do some adjustment of directory.

docs/design/environment/enviroment_management.md
healthjyk: Yes, this is allowed.

healthjyk:
There is new pull request https://github.com/KusionStack/kusion/pull/621, please track there, close this pull request.

PR #602, opened 2023-11-16
feat: add random string function for upcoming secret generator refact

adohe:
#### What type of PR is this?
/kind feature

add random string function for upcoming secret generator refact

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/generator/appconfiguration/generator/workload/secret/rand.go
+++ b/pkg/generator/appconfiguration/generator/workload/secret/rand.go
@@ -1,0 +1,52 @@
+package secret
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+	"unsafe"
+)
+
+// Since rand.NewSource() doesn't provide safety under concurrent use,
+// we need to use sync.Mutex here.
+var rng = struct {
+	sync.Mutex
+	rand *rand.Rand
+}{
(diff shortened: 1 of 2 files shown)


liu-hm19:
Where will this method be used? If a randomly generated string based on the current time is added to the rendering logic of application resources, it may potentially cause changes in the generated `Spec` or `Intent` with every `build`/`preview`/`apply` operation, which might have an impact on the 3-way diff process.

ffforest:
Is this only used for non-terraform resources? For terraform resources we should just use the `random` provider - provides same functionality and handles diff already.

adohe:
> Is this only used for non-terraform resources? To generate secrets in terraform resources we should just use the `random` provider - provides same functionality and handles diff already.

This is only used for secret generator, so we put this package under secret.

adohe:
> Where will this method be used? If a randomly generated string based on the current time is added to the rendering logic of application resources, it may potentially cause changes in the generated `Spec` or `Intent` with every `build`/`preview`/`apply` operation, which might have an impact on the 3-way diff process.

This method will be used to generate random token string if and only if developers don't provide one in configuration als the impact on diff logic is as expected.

elliotxx (approved):
lgtm

PR #610, opened 2023-11-21
feat: add external secret store api and interface

adohe:
#### What type of PR is this?
/kind feature

add external secret store api and interface

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/apis/secrets/types.go
+++ b/pkg/apis/secrets/types.go
@@ -1,0 +1,33 @@
+package secrets
+
+// SecretStoreSpec contains configuration to describe target secret store.
+type SecretStoreSpec struct {
+	Provider *ProviderSpec `yaml:"provider" json:"provider"`
+}
+
+// ProviderSpec contains provider-specific configuration.
+type ProviderSpec struct {
+	// AWS configures a store to retrieve secrets from AWS Secrets Manager.
+	AWS *AWSProvider `yaml:"aws,omitempty" json:"aws,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/apis/secrets/types.go
```diff
@@ -0,0 +1,33 @@
...
+// ProviderSpec contains provider-specific configuration.
+type ProviderSpec struct {
+	// AWS configures a store to retrieve secrets from AWS Secrets Manager.
+	AWS *AWSProvider `yaml:"aws,omitempty" json:"aws,omitempty"`
+	// Vault configures a store to retrieve secrets from HashiCorp Vault.
+	Vault *VaultProvider `yaml:"vault,omitempty" json:"vault,omitempty"`
+}
+
+// AWSProvider configures a store to retrieve secrets from AWS Secrets Manager.
+type AWSProvider struct {
```
SparkYuan, line 17: I am not sure whether the name `AWSProvider` is too general, considering this provider only supports one AWS product 'AWS Secrets Manager'

adohe: > I am not sure whether the name `AWSProvider` is too general, considering this provider only supports one AWS product 'AWS Secrets Manager'

good question, but I think we can put the problem on the back for now, environment management also include aws related configurations, we can go back this after @healthjyk gives a detailed api reference.

SparkYuan: okay

SparkYuan (approved):
lgtm

PR #614, opened 2023-11-23
feat: add vault secret store impl

adohe:
#### What type of PR is this?
/kind feature

add vault secret store impl

Fixes #

```release-note

```

```docs

```


pkg/secrets/providers/hashivault/fake/fake.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type (
+	ReadWithDataWithContextFn func(ctx context.Context, path string, data map[string][]string) (*vault.Secret, error)
+	Logical                   struct {
+		ReadWithDataWithContextFn ReadWithDataWithContextFn
+	}
+)
+
+func NewReadWithContextFn(secretData map[string]interface{}, err error) ReadWithDataWithContextFn {
```
SparkYuan, line 17: Is this fake file only used for tests? If so, it might be a good idea to move it to the test folder.

adohe: Package fake is a test-double implementation of `Logical`, it is useful for testing. Also I strongly recommend we follow this pattern in Kusion, this pattern can be considered as best practice.

SparkYuan (approved):
lgtm

PR #618, opened 2023-11-24
docs: add domain model of workspace management

### LINKED ISSUE #617: Feature: add workspace management domain models

Add workspace management domain models, including `WorkspaceConfig`, `Module`, `BackendConfig`, `RuntimeConfig`, etc.

### Why is this needed?

The first step to land workspace management.

healthjyk:
#### What type of PR is this?

/kind feature

add domain model of workspace management.

Fixes #617 

The package registry is used to workspace config deserialization.

### DIFF
--- a/pkg/models/config.go
+++ b/pkg/models/config.go
@@ -1,0 +1,46 @@
+package models
+
+// ModuleConfig is an interface to describe the config of a module, which is organized in the
+// ModuleConfigs to be provided in the workspace.Config.
+type ModuleConfig struct {
+	Module
+
+	// ProjectSelector is used to define the projects that consume the ModuleConfig.
+	ProjectSelector []string `yaml:"projectSelector,omitempty" json:"projectSelector,omitempty"`
+}
+
+// NewModuleConfig is used to new a ModuleConfig by a specified Module.
+func NewModuleConfig(module Module) *ModuleConfig {
+	return &ModuleConfig{Module: module}
+}
+
+// ModuleConfigs is a group of ModuleConfig.
+type ModuleConfigs map[string]*ModuleConfig
+
+// BackendConfig is an interface to describe the config of a backend.
+type BackendConfig interface {
+	// BackendName returns the name to identify the backend uniquely.
+	BackendName() string
+}
+
+// SecureBackendConfig is an interface to describe the verified config of a backend.
+type SecureBackendConfig interface {
+	BackendConfig
+
+	// Validate validates the BackendConfig is correct or not.
+	Validate() error
+}
+
+// RuntimeConfig is an interface to describe the config of a runtime.
+type RuntimeConfig interface {
+	// RuntimeName returns the name to identify the runtime uniquely.
+	RuntimeName() string
+}
+
+// SecureRuntimeConfig is an interface to describe the verified config of a runtime.
+type SecureRuntimeConfig interface {
+	RuntimeConfig
+
+	// Validate validates the RuntimeConfig is correct or not.
+	Validate() error
+}
--- a/pkg/models/config.go
+++ b/pkg/models/config.go
@@ -1,6 +1,6 @@
 package models
 
-// ModuleConfig is an interface to describe the config of a module, which is organized in the
+// ModuleConfig is a struct to describe the config of a module, which is organized in the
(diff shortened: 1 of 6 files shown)


pkg/models/module.go
```diff
@@ -0,0 +1,28 @@
...
+	ModuleName() string
+}
+
+// SecureWorkspaceModule is used to hold Module fields retrieved from workspace config which get
+// verified.
+type SecureWorkspaceModule interface {
+	Module
+
+	// ValidateWorkspaceModule validates the Module retrieved from workspace is correct or not.
+	ValidateWorkspaceModule() error
```
SparkYuan, line 18: Please rename `ValidateWorkspaceModule` to `Validate` to make it concise, considering it is already within the `SecureWorkspaceModule` definition.

pkg/models/module.go
```diff
@@ -0,0 +1,28 @@
+package models
+
+// Module is an interface to describe the domain model of a building block. A Module comes both
+// from an application developer-centric schema and workspace config, which can fully describes a workload,
+// an accessory, etc., such as database.Database, network.Port.
+// A Module can be built into a or a set of Resource, which will be consumed by Kusion engine.
+type Module interface {
+	// ModuleName returns the name to identify the module uniquely.
+	ModuleName() string
```
SparkYuan, line 9: Please rename `ModuleName` to `Name` to follow the Go convention

pkg/models/config.go
```diff
@@ -0,0 +1,46 @@
+package models
+
+// ModuleConfig is an interface to describe the config of a module, which is organized in the
```
SparkYuan, line 3: Whether ModuleConfig is an interface or a struct? The comment conflicts with the code

pkg/models/module.go
healthjyk: Cause the module has two sources: 1, scheme provided by application developer; 2, workspace config provided by platform engineer. To do validation separately, there are functions `ValidateWorkspaceModule` and `ValidateSchemeModule`.

pkg/models/module.go
healthjyk: The function `ModuleName` is used to define a module. If changed to `Name`, the capacity of definition will not exist.

pkg/models/config.go
healthjyk: It's a struct. This is a miswriting.

pkg/workspace/config.go
```diff
@@ -0,0 +1,38 @@
+package workspace
+
+import (
+	"kusionstack.io/kusion/pkg/models"
+)
+
+// Config is used to describe the config of a workspace.
+type Config struct {
```
adohe, line 8: Why not just use `type Workspace struct` here? What's the difference between `Config` and `Workspace`?

pkg/registry/registry.go
```diff
@@ -0,0 +1,32 @@
...
+// config structure, so that the workspace config can be correctly deserialized.
+package registry
+
+import (
+	"kusionstack.io/kusion/pkg/models"
+	"kusionstack.io/kusion/pkg/models/appconfiguration/accessories/database"
+)
+
+var (
+	registeredModules  []models.Module
```
adohe, line 11: The module code might be put out side of Kusion repo, so how to register modules here?

pkg/workspace/operator.go
```diff
@@ -0,0 +1,30 @@
+package workspace
```
adohe, line 1: Personally I would suggest you only include API types definition in this PR, operator code can be a separate PR.

healthjyk: That's cool and I'll change it.

pkg/registry/registry.go
healthjyk: The registry pattern is based on the assumption that Kusion can get all the module structure. For the module code  put outside kusion repo, the thing goes the same, kusion has to get the module structure, whatever the way.
Of course, there is more work, and the exact work is unknown. It depends on the new generator scheme.
Hence, it seems that forgetting about the registry, and make it simple is more suitable for now.

healthjyk:
This pr get closed, upgrade is https://github.com/KusionStack/kusion/pull/639

PR #639, opened 2023-11-30
feat: add workspace domain model

### LINKED ISSUE #617: Feature: add workspace management domain models

healthjyk:
#### What type of PR is this?

/kind feature

add domain model of workspace management.

Fixes #617 

For the `backend` config, should be completed later.

### DIFF
--- a/pkg/apis/workspace/modules.go
+++ b/pkg/apis/workspace/modules.go
@@ -1,0 +1,15 @@
+package workspace
+
+// ModuleConfigs is a set of multiple ModuleConfig, whose key is the module name.
+type ModuleConfigs map[string]ModuleConfig
+
+// ModuleConfig is the config of a module, which contains a default and several patcher blocks.
+//
+// The default block's key is "default", and value is the module inputs. The patcher blocks' keys
+// are the patcher names, which are just block identifiers without specific meaning, but must
+// not be "default". Besides module inputs, patcher block's value also contains a field named
+// "projectSelector", whose value is a slice containing the project names that use the patcher
+// configs. A project can only be assigned in a patcher's "projectSelector" field, the assignment
+// in multiple patchers is not allowed. For a project, if not specified in the patcher block's
+// "projectSelector" field, the default config will be used.
+type ModuleConfig map[string]GenericConfig
--- a/pkg/apis/workspace/util.go
+++ b/pkg/apis/workspace/util.go
@@ -1,5 +1,0 @@
-package workspace
-
-// GenericConfig is a generic model to describe config which shields the difference among multiple concrete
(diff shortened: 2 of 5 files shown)


pkg/apis/workspace/util.go
```diff
@@ -0,0 +1,5 @@
+package workspace
+
+// GenericConfig is a generic model to describe config which shields the difference among multiple concrete
+// models. GenericConfig is designed for extensibility, used for module, terraform runtime config, etc.
+type GenericConfig map[string]any
```
SparkYuan, line 5: Why is it preferable to define it in the util file? It would be more appropriate to rename `util` to `type`.

pkg/apis/workspace/workspace.go
```diff
@@ -0,0 +1,19 @@
+package workspace
+
+// Workspace is a logical concept representing a target that stacks will be deployed to.
+// Workspace is managed by platform engineers, which contains a set of configurations
+// that application developers do not want or should not concern, and is reused by multiple
+// stacks belonging to different projects.
+type Workspace struct {
+	// Name identifies a Workspace uniquely.
+	Name string `yaml:"-" json:"-"`
```
SparkYuan, line 9: Why should we ignore the `name` field in yaml and json, could you provide additional comments to clarify this?

healthjyk: This is because there is no `name` field in the configuration file

pkg/apis/workspace/modules.go
```diff
@@ -0,0 +1,15 @@
...
+// ModuleConfig is the config of a module, which contains a default and several patcher blocks.
+//
+// The default block's key is "default", and value is the module inputs. The patcher blocks' keys
+// are the patcher names, which are just block identifiers without specific meaning, but must
+// not be "default". Besides module inputs, patcher block's value also contains a field named
+// "projectSelector", whose value is a slice containing the project names that use the patcher
+// configs. A project can only be assigned in a patcher's "projectSelector" field, the assignment
+// in multiple patchers is not allowed. For a project, if not specified in the patcher block's
+// "projectSelector" field, the default config will be used.
+type ModuleConfig map[string]GenericConfig
```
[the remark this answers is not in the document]
healthjyk: I think you want `ModuleConfig` and has a fixed field named `Default`. I don't think this is a good idea. Here are the reasons:
1. If there is a field naming `Default` whose yaml key is "default", you have to do additional restriction that the patchers map should not have a key naming "default";
2. the default block should be on an equal footing to a patcher block, not the all patcher blocks.

PR #653, opened 2023-12-05
feat: getting workspace when running kusion build

### LINKED ISSUE #638: Feat: secret management supports Alicloud Secrets Manager as backend

Add Alicloud Secrets Manager as a backend in Kusion secret management. 

### Why is this needed?

Kusion needs to support various cloud secret managers, and Alicloud Secrets Manager is one of them.

### LINKED ISSUE #648: Feat: Using workspace config as "kusion build" input

When running kusion build, use workspace config as input.

### Why is this needed?

The kusion build input is separated into two parts: workspace and kcl schema, which the former is not supported for now.

healthjyk:
#### What type of PR is this?

/kind feature

using workspace when running kusion build

Fixes #648 

Fix #638 can start, and use it as a test case.

### DIFF
--- a/pkg/cmd/build/util.go
+++ b/pkg/cmd/build/util.go
@@ -15,12 +15,13 @@ import (
 	"kusionstack.io/kusion/pkg/apis/intent"
 	"kusionstack.io/kusion/pkg/apis/project"
 	"kusionstack.io/kusion/pkg/apis/stack"
+	apisworkspace "kusionstack.io/kusion/pkg/apis/workspace"
 	"kusionstack.io/kusion/pkg/cmd/build/builders"
 	"kusionstack.io/kusion/pkg/cmd/build/builders/kcl"
-	"kusionstack.io/kusion/pkg/modules/inputs"
-
 	"kusionstack.io/kusion/pkg/log"
+	"kusionstack.io/kusion/pkg/modules/inputs"
 	"kusionstack.io/kusion/pkg/util/pretty"
+	workspace "kusionstack.io/kusion/pkg/workspace"
 )
 
 func IntentWithSpinner(o *builders.Options, project *project.Project, stack *stack.Stack) (*intent.Intent, error) {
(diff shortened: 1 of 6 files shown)


pkg/modules/generators/app_configurations_generator.go
```diff
@@ -1,33 +1,38 @@
 package generators
 
 import (
+	"errors"
 	"fmt"
 
 	"kusionstack.io/kusion/pkg/apis/intent"
 	"kusionstack.io/kusion/pkg/apis/project"
 	"kusionstack.io/kusion/pkg/apis/stack"
+	apisworkspace "kusionstack.io/kusion/pkg/apis/workspace"
```
adohe, line 10: I would suggest rename `apisworkspace` to `workspaceapi`

pkg/modules/generators/app_configurations_generator.go
```diff
@@ -41,11 +46,19 @@ func NewAppConfigurationGenerator(
 		return nil, fmt.Errorf("can not find app configuration when generating the Intent")
 	}
 
+	if ws == nil {
+		return nil, errors.New("workspace must not be empty") // AppConfiguration asks for non-empty workspace
+	}
+	if err := workspace.ValidateWorkspace(ws); err != nil {
+		return nil, fmt.Errorf("invalid workspace %s, %w", stack.GetName(), err)
```
adohe, line 53: invalid workspace config

pkg/cmd/build/util.go
```diff
@@ -117,6 +125,24 @@ func buildAppConfigs(o *builders.Options, stack *stack.Stack) (map[string]inputs
...
 
+func getWorkspace(stackName string) (*apisworkspace.Workspace, error) {
+	wsOperator, err := workspace.NewDefaultOperator()
+	if err != nil {
+		return nil, fmt.Errorf("new workspace operator failed, %w", err)
+	}
+	// stack name should be same as the workspace name
+	ws, err := wsOperator.GetWorkspace(stackName)
+	// allow empty workspace
+	if errors.Is(err, workspace.ErrWorkspaceNotExist) {
```
adohe, line 136: what will happen if workspace is nil? Seems `AppConfigurationGenerator` returns err if workspace is nil

healthjyk: There is no limit of the empty workspace is because if there are other app configuration schemas which are developed by kusion's users, the workspace may not be necessary. (Just consider an independent developer, maybe he/she will declare all his/her configurations in a new app configuration schema.) Hence, in pkg cmd, if the workspace is empty, there used to be no error returned.
But after revisiting the code, the operation happens under the condition using `appConfigurationBuilder`. It means using kusion‘s own `AppConfiguration` schema, which asks for non-nil workspace. So I should delete the special error handling. You give me a great vetting, thanks!

pkg/modules/generators/app_configurations_generator.go
healthjyk: I am used to add the directory one level up as prefix if there are packages with same name, to keep the same alias rule.
But `workspaceapi` seems clearer than `apisworkspace`, I'll update it.

PR #662, opened 2023-12-07
feat: add mysql/oss/s3 as new backends

### LINKED ISSUE #647: Feat: Update workspace backend config and use the workspace backend config for state

1. In workspace config, support the backends that state now use, including: mysql(db), oss, s3; 
2. support using backend config from workspace as the state backend config;
3. update state backend config, such as: rename database to mysql, local config cannot specify the file, etc.

### Why is this needed?

The above jobs have not done.

healthjyk:
#### What type of PR is this?

/kind feature

add mysql/oss/s3 as new backends

Fixes #647


SparkYuan (approved):
lgtm

PR #668, opened 2023-12-11
feat: update workspace util func and fix known bug

healthjyk:
…ce dir is not hidden folder

#### What type of PR is this?

/kind feature

1. add CURD func by default workspace operator;
2. use inline for the `GenericObjectStorageConfig` for `OssConfig` and `S3Config`;
3. correct validation for `S3Config`.

### DIFF
--- a/pkg/workspace/operator.go
+++ b/pkg/workspace/operator.go
@@ -28,23 +28,66 @@ var (
 	ErrWorkspaceAlreadyExist = errors.New("workspace has already existed")
 )
 
+// CheckWorkspaceExistenceByDefaultOperator checks the workspace exists or not by default operator.
+func CheckWorkspaceExistenceByDefaultOperator(name string) (bool, error) {
+	operator, err := NewValidDefaultOperator()
+	if err != nil {
+		return false, err
+	}
+	return operator.WorkspaceExist(name), nil
+}
+
+// GetWorkspaceByDefaultOperator gets a workspace by default operator.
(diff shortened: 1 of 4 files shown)


SparkYuan:
What is the `pkg/workspace/testdata/invalid_workspaces_dir/dir_workspace_file/.gitkeep` file used for?

pkg/workspace/operator.go
```diff
@@ -28,6 +28,60 @@ var (
...
+	operator, err := NewValidDefaultOperator()
+	if err != nil {
+		return err
+	}
+	return operator.DeleteWorkspace(name)
+}
+
 // Operator is used to handle the CURD operations of workspace. Operator only supports local file
 // system as backend for now.
 type Operator struct {
```
SparkYuan, line 87: Why not define the Operator as an interface as we will have various types of operators in the future?

healthjyk: That's a good suggestion! But it's for the future, now it's more visualized.
Now it seems C/U/R/D defines the functions of the interface "Operator", but if there is any more functions need, we can get more feedback from may-exist users and subsequent R&D work.
Besides, even though the operator becomes an interface, there still needs a struct holding the structure of the configuration of the Operator, which may have two field `backend type` and `backend config`, and it's a struct.
Defining Operator an interface is a good idea, but I think we can do it when there are more workspace operator backends, now the struct is ok. What's your opinion?

SparkYuan (approved):
lgtm

PR #708, opened 2023-12-20
feat: add func of getting specified type from workspace genericConfig

healthjyk:
#### What type of PR is this?

/kind feature

add func of getting specified type from workspace genericConfig.

### DIFF
--- a/pkg/workspace/util.go
+++ b/pkg/workspace/util.go
@@ -5,22 +5,13 @@ import (
 	"fmt"
 	"os"
 
-	"kusionstack.io/kusion/pkg/apis/core/v1"
+	v1 "kusionstack.io/kusion/pkg/apis/core/v1"
 )
 
-var (
-	ErrEmptyProjectName          = errors.New("empty project name")
-	ErrEmptyModuleConfigs        = errors.New("empty module configs")
-	ErrEmptyProjectModuleConfigs = errors.New("empty module configs of the project")
-	ErrEmptyProjectModuleConfig  = errors.New("empty module config of the project")
+var ErrEmptyProjectName = errors.New("empty project name")
 
-	ErrEmptyRuntimeConfigs   = errors.New("empty runtime configs")
-	ErrEmptyKubernetesConfig = errors.New("empty kubernetes config")
-	ErrEmptyTerraformConfig  = errors.New("empty terraform config")
-)
-
-// CompleteWorkspace sets the workspace name and default value of unset item, should be called after ValidateWorkspace.
-// The config items set as environment variables are not got by CompleteWorkspace.
+// CompleteWorkspace sets the workspace name and default value of unset item, should be called after Validatev1.
+// The config items set as environment variables are not got by Completev1.
 func CompleteWorkspace(ws *v1.Workspace, name string) {
 	if ws.Name != "" {
 		ws.Name = name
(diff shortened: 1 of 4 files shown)


pkg/workspace/util.go
```diff
@@ -237,3 +211,63 @@ func CompleteWholeS3Config(config *workspace.S3Config) {
 		config.Region = region
 	}
 }
+
+// GetIntFieldFromGenericConfig returns the value of the key in config which should be of type int.
+// If exist but not int, return error. If not exist, return 0, nil.
+func GetIntFieldFromGenericConfig(config workspace.GenericConfig, key string) (int, error) {
```
adohe, line 217: GetIntFieldFromGenericConfig -> GetIntFromGenericConfig

adohe, line 217: compared to return error, return yes or no might be better?

pkg/workspace/util.go
```diff
@@ -237,3 +211,63 @@ func CompleteWholeS3Config(config *workspace.S3Config) {
...
+	value, ok := config[key]
+	if !ok {
+		return 0, nil
+	}
+	i, ok := value.(int)
```
adohe, line 222: how about using `int32` here?

pkg/workspace/util.go
```diff
@@ -237,3 +211,63 @@ func CompleteWholeS3Config(config *workspace.S3Config) {
...
+	if !ok {
+		return 0, fmt.Errorf("the value of %s is not map", key)
+	}
+	return i, nil
+}
+
+// GetStringFieldFromGenericConfig returns the value of the key in config which should be of type string.
+// If exist but not string, return error; If not exist, return "", nil.
+func GetStringFieldFromGenericConfig(config workspace.GenericConfig, key string) (string, error) {
```
adohe, line 231: GetStringFieldFromGenericConfig -> GetStringFromGenericConfig

pkg/workspace/util.go
```diff
@@ -237,3 +211,63 @@ func CompleteWholeS3Config(config *workspace.S3Config) {
...
+	s, ok := value.(string)
+	if !ok {
+		return "", fmt.Errorf("the value of %s is not string", key)
+	}
+	return s, nil
+}
+
+// GetMapFieldFromGenericConfig returns the value of the key in config which should be of type map[string]any.
+// If exist but not map[string]any, return error; If not exist, return nil, nil.
+func GetMapFieldFromGenericConfig(config workspace.GenericConfig, key string) (map[string]any, error) {
```
adohe, line 245: GetMapFieldFromGenericConfig -> GetMapFromGenericConfig

pkg/workspace/util.go
```diff
@@ -237,3 +211,63 @@ func CompleteWholeS3Config(config *workspace.S3Config) {
...
+	m, ok := value.(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("the value of %s is not map", key)
+	}
+	return m, nil
+}
+
+// GetStringMapFieldFromGenericConfig returns the value of the key in config which should be of type map[string]string.
+// If exist but not map[string]string, return error; If not exist, return nil, nil.
+func GetStringMapFieldFromGenericConfig(config workspace.GenericConfig, key string) (map[string]string, error) {
```
adohe, line 259: GetStringMapFieldFromGenericConfig -> GetStringMapFromGenericConfig

pkg/workspace/util.go
healthjyk: If return `int32`, there should be a new func named `GetInt32FromGenericConfig` cause the returned type is fixed.

pkg/workspace/util.go
healthjyk: If return yes or no, when using these functions, the calling one always still need to build an error. Besides, when calling `GetMapFromGenericConfig`, there are two kinds of errors, the first is `not map`, the second is `field is not string`, using `bool` is not enough.

healthjyk: fixed

pkg/workspace/util.go
healthjyk: fixed

pkg/workspace/util.go
healthjyk: fixed

pkg/workspace/util.go
healthjyk: fixed

liu-hm19 (approved):
lgtm

PR #716, opened 2023-12-22
fix: misuse of env KUSION_PATH

healthjyk:
#### What type of PR is this?

/kind bug

Use the value of env `KUSION_HOME` instead of `KUSION_PATH` as the path to store kusion data, cause the former is `$HOME/.kusion` and the latter is `$HOME/.kusion/bin`. This change happens when changing the catalog of kusion package.


adohe (approved):
/lgtm

PR #717, opened 2023-12-22
feat: use workspace config when build service and job

### LINKED ISSUE #666: Feat: refactor the workload model according to the new design doc

- [ ] Delete the non-app-oriented fields in the workload schema
- [ ] Refactor the Workload generator according to the new design doc

### Why is this needed?

We already have updated our directory structure in the latest design doc, so we should update related models.

healthjyk:
#### What type of PR is this?

/kind feature

Use workspace config when build service and job.

Fixes #666

### DIFF
--- a/pkg/modules/generators/workload/workload_generator.go
+++ b/pkg/modules/generators/workload/workload_generator.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/imdario/mergo"
 	"golang.org/x/exp/maps"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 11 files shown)


pkg/modules/generators/workload/workload_generator.go
```diff
@@ -405,3 +410,36 @@ func handleFileCreation(c container.Container, uniqueAppName, containerName stri
...
+	}
+	if base.Replicas == 0 {
+		base.Replicas = replicas
+	}
+	labels, err := workspace.GetStringMapFromGenericConfig(config, workload.FieldLabels)
+	if err != nil {
+		return err
+	}
+	if labels != nil {
+		if err = mergo.Merge(&base.Labels, labels); err != nil {
```
liu-hm19, line 431: How does `mergo.Merge` process two maps with the same key? According to our proposal, we are supposed to assure that the labels and annotations from developers won't be overridden by workspace configs.

liu-hm19 (approved):
lgtm

PR #748, opened 2024-01-11
feat: update kusion init, and reuse project name flag

healthjyk:
#### What type of PR is this?

/kind feature

update kusion init, and reuse project name flag.


healthjyk:
avoid recurring of flag projectName

pkg/scaffold/internal/single-stack-sample/dev/kcl.mod
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "{{ .ProjectName }}"
-edition = "0.5.0"
+edition = "0.1.0"
```
SparkYuan, line 3: Does this edition represent the version of KCL? As far as I know, the KCL version we used is not 0.1.0

ffforest (dismissed):

pkg/scaffold/internal/single-stack-sample/dev/kcl.mod
healthjyk: do not need edition anymore

liu-hm19 (approved):

PR #803, opened 2024-02-19
feat: add code runner for configuration code execution

adohe:
#### What type of PR is this?
/kind feature

add code runner to execute configuration code

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/generate/run/fake/fake.go
+++ b/pkg/cmd/generate/run/fake/fake.go
@@ -1,0 +1,13 @@
+package fake
+
+import "kusionstack.io/kusion/pkg/cmd/generate/run"
+
+var _ run.CodeRunner = &KPMRunner{}
+
+// KPMRunner is a fake code runner for testing purposes.
+type KPMRunner struct{}
+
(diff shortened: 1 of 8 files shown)


liu-hm19 (approved):
lgtm

pkg/cmd/generate/run/fake/fake.go
```diff
@@ -0,0 +1,13 @@
+package fake
```
healthjyk, line 1: Maybe you should move all these codes to an isolate package named `generator` or something else. The package `cmd` should be as thin as possible. 
Besides, the package seems tricky, is it from TDD or something else?

adohe: Code snippets in the `run` package are only used by `kusion generate` command, so we put all it under the `generate` package, the `fake` package contains fake code runner which is necessary for testing, besides this is not a tricky design, the fake mechanism is wildly used in Kubernetes.

healthjyk (approved):

PR #804, opened 2024-02-20
feat: add crd resource visitor util

adohe:
#### What type of PR is this?
/kind feature

add crd resource visitor util

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/util/resource/crd/multi_crd.yaml
+++ b/pkg/cmd/util/resource/crd/multi_crd.yaml
@@ -1,0 +1,51 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.4
+  creationTimestamp: null
+  name: guests.webapp.my.domain
+spec:
+  group: webapp.my.domain
+  names:
+    kind: Guest
+    listKind: GuestList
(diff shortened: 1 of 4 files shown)


pkg/cmd/util/resource/visitor.go
```diff
@@ -0,0 +1,64 @@
+package resource
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"k8s.io/apimachinery/pkg/util/yaml"
```
healthjyk, line 9: There are multiple yaml packages in kusion, using the same is better.
And why you use the yaml official package "gopkg.in/yaml.v3"

pkg/cmd/util/resource/crd/multi_crd.yaml
```diff
@@ -0,0 +1,51 @@
+
+---
```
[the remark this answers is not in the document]
adohe: Add `---` to simulate multi YAML documents within one YAML, so this is necessary.

pkg/cmd/util/resource/visitor.go
adohe: crd resources might be in YAML or JSON format, for the crd resource visitor, we have to take this into consideration, and the yaml package in `apimachinery` provides convenient method to parse YAML or JSON documents.

healthjyk (approved):

PR #814, opened 2024-02-23
fix: use package path/filepath instead of path to support windows

### LINKED ISSUE #815: Bug: kusion workspace create failed on windows
### What happened?

Executing `kusion workspace create` failed on windows, with error "open file failed"

Executing `kusion workspace create` succeed on windows.

healthjyk:
#### What type of PR is this?

/kind bug

Provide correct path operation under windows. 

Fixes #815


liu-hm19 (approved):
lgtm

PR #877, opened 2024-03-06
feat: add workspace edit cmd

### LINKED ISSUE #867: Feat: use the default workspace config when the specified workspace is not found

The user-specified workspaces (e.g. `dev`, `pre` or `prod`) currently need to exist to perform the related operation like `kusion generate` and `kusion apply`, please add support for the `default` workspace so that kusion can use the default workspace config (`default.yaml`) when the user-specified workspace is not found. 

### Why is this needed?

This can bring more convenience for the kusion cli users.

liu-hm19:
#### What type of PR is this?

/kind feature
This PR adds the implementation of `kusion workspace edit` cmd. 

Fixes #867 

```release-note

```

```docs

```


liu-hm19:
The implementation of this feature is temporarily deferred, so we are closing this PR.

PR #902, opened 2024-03-11
feat: add state storage implementation: local, oss, s3, mysql

healthjyk:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/apis/core/v1/state.go
+++ b/pkg/apis/core/v1/state.go
@@ -40,7 +40,7 @@ type State struct {
 	CreateTime time.Time `json:"createTime" yaml:"createTime"`
 
 	// ModifiedTime is the time State is modified each time
-	ModifiedTime time.Time `json:"modifiedTime,omitempty" yaml:"modifiedTime"`
+	ModifiedTime time.Time `json:"modifiedTime,omitempty" yaml:"modifiedTime,omitempty"`
 }
 
 func NewState() *State {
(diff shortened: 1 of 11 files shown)


healthjyk:
Unify the behaviors across multiple state storages. Compared to the old state storage, there are part of the bug fixes and functional escalation:
- Delete otiose function `Delete()` of interface `StateStorage`;
- unify the behavior of function `Apply()`, only store the state to backend. While the old local storage updates modifiedTime, while others not.
- Unify the storage requirements for all state storage, which asks for no repetitive project name. While for old state storages, only the default local storage allows repetitive project name (state file under stack directory), while others not, even the local storage with specified path does not support. To support projects with the same name, there will be another issue to track it.
- Refactor the default local state storage to ~/.kusion/states/{projectName}/{stackName}/{workspaceName}/state. yaml
- Upgrade mysql table to (project, stack, workspace, content), where the content is the yaml.Marshal of state. This table structure is to fit in the changeable the state structure, take ref of Terraform. While for old mysql storage, do not update `id` field, which is not correct, cause not all mysql db has auto-increasing id.
- Unify the behavior of oss and s3 when state file not exist, return nil with no error, keep the same with others. While the old oss/s3 storage return error. And delete useless list operation.
- StateStorage works for a specified project-stack-workspace, not  for multiple.
- Update the State model, delete field "Tenant", replace "Cluster" with "Workspace", ref: https://github.com/KusionStack/kusion/pull/897
@adohe @liu-hm19 @SparkYuan

liu-hm19 (approved):
lgtm

PR #942, opened 2024-03-19
feat: update workspace cmds, add backend and current flags

healthjyk:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/workspace/create/options.go
+++ b/pkg/cmd/workspace/create/options.go
@@ -3,13 +3,15 @@ package create
 import (
 	"fmt"
 
+	"kusionstack.io/kusion/pkg/backend"
 	"kusionstack.io/kusion/pkg/cmd/workspace/util"
-	"kusionstack.io/kusion/pkg/workspace"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/cmd/workspace/create/options.go
```diff
@@ -26,7 +28,7 @@ func (o *Options) Complete(args []string) error {
 }
 
 func (o *Options) Validate() error {
-	if err := util.ValidateName(o.Name); err != nil {
+	if err := util.ValidateNotDefaultName(o.Name); err != nil {
```
liu-hm19, line 31: If users are not allowed to create the `default` workspace with `kusion workspace create`, then how should they init a `default` workspace for a new backend? Will we provide relevant cmd?

healthjyk: The default workspace will get created when you run `kusion workspace/apply/preview/destroy/build`, that is when you use workspace.

liu-hm19 (approved):
lgtm

PR #956, opened 2024-03-25
feat: update yaml package for appconfiguration generator

### LINKED ISSUE #940: Bug: using yaml.v2 for deserialization may lead to failure in json serialization
### What happened?

Since the `Env` in `Container` of the `Workload` is the type of `MapSlice`, which only exists in `yaml.v2` package, currently we use `yaml.v2` for unmarshaling  module resources. But it does not support the nested `map[string]interface{}`, resulting it become `map[interface{}]interface{}`, which may easily cause the subsequent JSON marshaling to fail because JSON keys can only be the type of `string`. 

Here is a related discussion on GitHub:  https://github.com/go-yaml/yaml/issues/139

image

Please unify the usage of `yaml.v2` & `yaml.v3`. For me, better to use `yaml.v3`. 

Execute `kusion build`, `kusion preview`, `kusion apply` or `kusion destroy` on Kusion module with nested `map[string]interface{}` config. 

_No response_

### Kusion version

<details>

```console
$ kusion version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

### Install tools

<details>

</details>

liu-hm19:
#### What type of PR is this?

/kind feature
This updates the use of the yaml package in app_configurations_generator, unmarshaling `Workload` with **yaml.v2** and unmarshaling other generator request attributes with **yaml.v3**. 

Fixes #940 

```release-note

```

```docs

```

### DIFF
--- a/pkg/modules/generators/app_configurations_generator.go
+++ b/pkg/modules/generators/app_configurations_generator.go
@@ -5,11 +5,12 @@ import (
 	"errors"
 	"fmt"
 
-	"gopkg.in/yaml.v2"
+	yamlv2 "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 
-	"kusionstack.io/kusion/pkg/apis/core/v1"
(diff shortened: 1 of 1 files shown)


SparkYuan (approved):
lgtm

PR #1052, opened 2024-04-18
fix: unexpected output in format byte slice

healthjyk:
#### What type of PR is this?

/kind bug

Fix unexpected spec output format in byte slice.

image


adohe (approved):
/lgtm

PR #1067, opened 2024-04-24
feat: export some plugin functions

SparkYuan:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/modules/generators/app_configurations_generator.go
+++ b/pkg/modules/generators/app_configurations_generator.go
@@ -134,7 +134,7 @@ func (g *appConfigurationGenerator) Generate(spec *v1.Spec) error {
 
 	// patch workload with resource patchers
 	for _, p := range patchers {
-		if err = patchWorkload(&wl, &p); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/modules/generators/app_configurations_generator.go
```diff
@@ -134,7 +134,7 @@ func (g *appConfigurationGenerator) Generate(spec *v1.Spec) error {
 
 	// patch workload with resource patchers
 	for _, p := range patchers {
-		if err = patchWorkload(&wl, &p); err != nil {
```
Liuxingyu1111111, line 137: Is the interface defined as multi-resource and patcher a better implementation?

SparkYuan: We will provide a new function to support patching other resources  in the next version

Liuxingyu1111111: ok

Liuxingyu1111111 (approved):

PR #1175, opened 2024-06-21
fix: skip replacing secret data without ref prefix

liu-hm19:
#### What type of PR is this?

/kind bug
This PR fixes a bug of replacing secret ref prefix in `kusion apply` command, which skips the secret data without ref prefix. 

Fixes #

```release-note

```

```docs

```


Liuxingyu1111111 (approved):
lgtm

PR #1179, opened 2024-06-24
feat: add support for on-premises secret provider define in api.kusion.io

### LINKED ISSUE #1177: Feat: refactor the structure for secret provider spec to support adding on-premises secret provider without adding new provider define in api.kusion.io

refactor the structure for secret provider spec to support add secret provider without add new provider define in api.kusion.io

### Why is this needed?

Now if I want to add new secret provider in specific scene(some secret provider is only use in some company thus it can not be open source),  I need to add new provider spec in apis/api.kusion.io/v1. I think apis/api.kusion.io/v1 should be consistent among different use case

Liuxingyu1111111:
#### What type of PR is this?

/kind feature

Now if I want to add new secret provider in specific scene(some secret provider is only use in some company thus it can not be open source), I need to add new provider spec in apis/api.kusion.io/v1. I think apis/api.kusion.io/v1 should be consistent among different use case

Fixes #1177

```release-note

```

```docs

```

### DIFF
--- a/pkg/apis/api.kusion.io/v1/types.go
+++ b/pkg/apis/api.kusion.io/v1/types.go
@@ -721,6 +721,9 @@ type ProviderSpec struct {
 
 	// Fake configures a store with static key/value pairs
 	Fake *FakeProvider `yaml:"fake,omitempty" json:"fake,omitempty"`
+
+	// Onprem configures a store in on-premises environments
+	OnPremises *OnPremisesProvider `yaml:"onpremises,omitempty" json:"onpremises,omitempty"`
 }
 
 // AlicloudProvider configures a store to retrieve secrets from Alicloud Secrets Manager.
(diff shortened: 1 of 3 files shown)


pkg/apis/api.kusion.io/v1/types.go
```diff
@@ -721,6 +721,9 @@ type ProviderSpec struct {
 
 	// Fake configures a store with static key/value pairs
 	Fake *FakeProvider `yaml:"fake,omitempty" json:"fake,omitempty"`
+
+	// Onprem configures a store in on-premises environments
+	OnPremises *OnPremisesProvider `yaml:"onprem,omitempty" json:"onprem,omitempty"`
```
SparkYuan, line 726: Please rename the filed name to align with the yaml/json tag

pkg/secrets/providers.go
```diff
@@ -85,6 +85,9 @@ func (ps *Providers) getProviderByName(providerName string) (SecretStoreProvider
 }
 
 func getProviderName(spec *v1.ProviderSpec) (string, error) {
+	if spec.OnPremises != nil && spec.OnPremises.Name != "" {
+		return spec.OnPremises.Name, nil
```
SparkYuan, line 89: What if `spec.OnPremises` is not null but `spec.OnPremises.Name` is null

SparkYuan (approved):
lgtm

PR #1211, opened 2024-07-09
feat: support runtime configurations in workspace

### LINKED ISSUE #1192: Feat: support Kubernetes multi-cluster configuration

Please support the declaration of Kubernetes runtime configurations in **Workspace** to manage multi-clusters with different workspace configurations.

The configuration may includes: 

- the local file path of `KubeConfig`
- the URL for obtaining `KubeConfig` content, e.g. S3 endpoint
- the `KubeConfig` content itself

## Background

Kusion needs to deploy the application resources to a specific Kubernetes cluster, thus it should allow users to specify the information of the targeted cluster. Currently, users can specify the cluster by configuring the `KUBECONFIG` environment variable, which stay consistent with `kubectl`. The related codes can be found [here](https://github.com/KusionStack/kusion/blob/main/pkg/engine/runtime/kubernetes/kubeops/config.go#L27)

## Target

In order to better integrate Kusion in the CI/CD pipeline and support the multi-cluster scenarios more flexibly, we hope to add Kubernetes runtime configurations in Kusion **Workspace**, for example, it can include the following configs: 

- the local file path of `KubeConfig`
- the URL for obtaining `KubeConfig` content, e.g. S3 endpoint
- the `KubeConfig` content itself

So that each workspace will correspond to a separate Kubernetes cluster, and users don't need to re-set the `KUBECONFIG` environment variable every time before deploying to a different Kubernetes cluster. 

## Main Concern

Actually, we once supported declaring Kubernetes and Terraform runtime configurations in Workspace in previous versions, but later we removed it for the following reasons: 

- Workspace is accessible to many team members, which may easily lead to the leakage of sensitive information, such as K8s cluster certificates and TF provider AK/SK. 
- `KubeConfig` can usually be considered as a workspace-level configuration, but the config of Terraform Provider is very like to be at the resource level, which means that the resources of different modules in the same workspace may still differ

Currently, we also support specifying the runtime information in the `Extensions` field of `Resource` in `Spec`. **Now, we need to consider the actual requirements and situations comprehensively to implement this feature.**

liu-hm19:
#### What type of PR is this?

/kind feature
This PR adds supports for runtime configurations in workspace. 

Fixes #1192 

```release-note

```

```docs

```

### DIFF
--- a/pkg/apis/api.kusion.io/v1/types.go
+++ b/pkg/apis/api.kusion.io/v1/types.go
@@ -865,7 +865,7 @@ type Spec struct {
 	// SecretSore represents a external secret store location for storing secrets.
 	SecretStore *SecretStore `yaml:"secretStore" json:"secretStore"`
 	// Context contains workspace-level configurations, such as runtimes, topologies, and metadata, etc.
-	Context GenericConfig `yaml:"context,omitempty" json:"context,omitempty"`
+	Context GenericConfig `yaml:"context" json:"context"`
 }
 
 // State is a record of an operation's result. It is a mapping between resources in KCL and the actual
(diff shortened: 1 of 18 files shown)


SparkYuan (approved):
lgtm

PR #1260, opened 2024-08-06
feat: remove requirement to use imported resource when apply

### LINKED ISSUE #1259: Feat: remove requirement to use imported resource when apply

- [x] Remove the requirement to use imported resources declared in the workspace every `kusion apply`.
- [x] ApplyPreview or Dryrun of imported resource change to `terraform import` instead of `terraform plan`. 

### Why is this needed?

- Currently, when we add a resource to the workspace with `importedResources`, every time we `kusion apply` a project in that workspace, we will be required to use that or every imported resources. This will cause some projects that do not use imported resources to be unable to apply.
- One more problem is when applying, the terraform resource that was declared in the workspace, in the phase generate from .k to spec the attributes of it will still receive data by kusion catalog so when we import, the action of it becomes `Update` not `UnChanged`. So the resource will update unnecessary fields.

hoangndst:
#### What type of PR is this?
- Currently, when we add a resource to the workspace with importedResources, every time we kusion apply a project in that workspace, we will be required to use that or every imported resources. This will cause some projects that do not use imported resources to be unable to apply.
- One more problem is when applying, the terraform resource that was declared in the workspace, in the phase generate from .k to spec the attributes of it will still receive data by kusion catalog so when we import, the action of it becomes Update not UnChanged. So the resource will update unnecessary fields.

/kind feature

Fixes #1259 

```release-note

```

```docs

```

### DIFF
--- a/pkg/engine/runtime/terraform/terraform_runtime.go
+++ b/pkg/engine/runtime/terraform/terraform_runtime.go
@@ -63,6 +63,27 @@ func (t *Runtime) Apply(ctx context.Context, request *runtime.ApplyRequest) *run
 
 	// dry run by terraform plan
 	if request.DryRun {
+		// the import resource should be import and return the resource from tfstate
+		if _, ok := plan.Extensions[tfops.ImportIDKey].(string); ok {
+			readResponse := t.Read(ctx, &runtime.ReadRequest{
+				PriorResource: request.PriorResource,
+				PlanResource:  request.PlanResource,
+				Stack:         request.Stack,
+			})
+			if readResponse.Status != nil {
+				return &runtime.ApplyResponse{Resource: &apiv1.Resource{}, Status: readResponse.Status}
+			}
+			return &runtime.ApplyResponse{
+				Resource: &apiv1.Resource{
+					ID:         plan.ID,
+					Type:       plan.Type,
+					Attributes: readResponse.Resource.Attributes,
+					DependsOn:  plan.DependsOn,
+					Extensions: plan.Extensions,
+				},
+				Status: nil,
+			}
+		}
 		pr, err := ws.Plan(ctx)
 		if err != nil {
 			return &runtime.ApplyResponse{Resource: nil, Status: v1.NewErrorStatus(err)}
--- a/pkg/modules/generators/app_configurations_generator.go
+++ b/pkg/modules/generators/app_configurations_generator.go
@@ -663,8 +663,8 @@ func patchImportedResources(resources v1.Resources, projectImportedResources map
 	for kusionID, importedID := range projectImportedResources {
 		if res, ok := resIndex[kusionID]; ok {
 			res.Extensions[tfops.ImportIDKey] = importedID
-		} else {
-			return fmt.Errorf("empty kusion id: %s", kusionID)
+			// remove the resource attribute to avoid update conflict when using terraform import
+			res.Attributes = make(map[string]interface{})
 		}
 	}
 


liu-hm19:
@hoangndst Hi, sorry for taking some time to review this PR. We discussed about the support of Kusion for importing existing resources again. 

Regarding the issue of `empty kusion id`, actually there is a field called `projectSelector` in the `modules` configurations in `Workspace`, where users can declare the specific application names that the module configurations applied to, which can help to avoid affecting unrelated applications when importing existing cloud resources. You can find more details [here](https://www.kusionstack.io/docs/concepts/workspace#modules). 

And regarding the issue of remote imported resources being updated by the current configurations during `kusion apply`, I believe the code modifications in this PR is quite reasonable, but I am a little concerned that clearing the `attributes` field of the imported resources might cause some unexpected problems. **I will validate with some test cases ASAP, and I will merge this PR after the validation**. 

In addition, we think that there should still be users who hope to use Kusion to update the imported resources, such as taking over the applications previously created with Terraform or cloud vendor console. Therefore, we are considering whether to add a field in the `modules` configurations in `Workspace` to act as a switch for updating the imported resources according to the current configurations. And we would like to hear your opinions : ) 

cc @ffforest

hoangndst:
About clearing the `attributes` field of imported resource is about ignore the update conflict. It just re-import to refresh update data of resource, not try to update change to make phase to `Update`. This also help we don't have to modify code to remove the `Update` phase if object is import resource.

The issue of empty kusion id, actually there is a field called projectSelector in the modules configurations in Workspace. About this I thinks for more flexible, we still need to remove requirement to use imported resource. More example: 2 project declared in projectSelector, and this `patch` block has 2 imported resource. One of projects just need to use 1 imported resource.

The case use Kusion to update the imported resources, such as taking over the applications previously created with Terraform or cloud vendor console. I agree with you for that case and add a field in the modules configurations in Workspace to act as a switch for updating the imported resources is a good idea. 

For more I think if will be more helpful if we add a field in modules configurations just for imported resource, all resources of project  in workspace that created will be added in that fields. For example: in case we created a project in that workspace and in other project in the same workspace where we also want to use the resources created in the previous project, we must import again. Only one concerned that about the config to allow which project to use a specific resource.

Looking for you guys opinions @liu-hm19 @ffforest

liu-hm19:
@hoangndst Thanks for sharing your constructive ideas! I agree with you that we can remove the validation for `empty kusion id`, which can make it more flexible. 

Regarding clearing the `Attributes` of the imported resources, our main concern is whether the resulting `Spec` can be normally applied and whether the rollback feature we are planning to implement in the future can work properly with the corresponding `State`. I'm doing some validations, and once the test cases are passed through, I will merge this PR : ) 

As for the feature of reusing the resources applied by an application with Kusion, I think this is indeed an exciting idea. And we can further discuss this later together with the field used to declare an imported resource to be updated. In fact, besides the way to add a field in `Workspace` to achieve this, we have also planned a `dependency` field in `AppConfiguration`, which might also be used for reusing the resources. The schema of `AppConfiguration` can be found [here](https://www.kusionstack.io/docs/concepts/app-configuration).

liu-hm19:
@hoangndst I've tested this PR and found that it works fine with the remote cloud resources, but for local Terraform resources like `random_password`, some fields of the `state` generated after executing `terraform import` will change each time. For example, you can try the following Terraform codes and run `terraform import random_password.wordpress-demo-mysql cXIjA_qrAyb8guT2`, and you should find that the `bcrypt_hash` of state imported each time is different. This will cause Kusion to calculate the operation type as `Update` instead of `Unchanged`, resulting in applying an empty resource and causing errors. 

main.tf.json: 
```
{
  "import": {
    "id": "cXIjA_qrAyb8guT2",
    "to": "random_password.wordpress-demo-mysql"
  },
  "provider": {
    "random": {}
  },
  "resource": {
    "random_password": {
      "wordpress-demo-mysql": {}
    }
  },
  "terraform": {
    "required_providers": {
      "random": {
        "source": "registry.terraform.io/hashicorp/random",
        "version": "3.6.0"
      }
    }
  }
}
```

However, I think this corner case might be a problem with the process logic of applying the imported resources. Perhaps we need to set the `Attributes` of the `prior` and `planned` resource the same as the `live` resource when applying the imported resources. Therefore, I will merge this PR first, and later create an issue to track this problem and fix it. Please feel free to continue pay attention on this feature and welcome to contribute : ) Thank you! 

cc @ffforest

liu-hm19 (approved):
lgtm

PR #1274, opened 2024-08-21
fix: json patch allows empty path on add and remove

### LINKED ISSUE #1249: Bug: json patch with patcher failed when adding custom volume to k8s yaml with no volume define in workload yaml
### What happened?

When I want to add volume to workload yaml with a kusion module, it return the error below 
`apply json patch to:***(resource id of workload)failed with error add operation does not apply: doc is missing path: "/spec/template/spec/volumes/-": missing value`

support adding custom volume to k8s yaml with no volume define in workload yaml

generate a workload yaml with no volume and try to patch a custom volume with another kusion module

_No response_

### Kusion version

<details>

```console
$ kusion version
v0.12.0
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

### Install tools

<details>

</details>

liu-hm19:
#### What type of PR is this?

/kind bug
This PR allows missing path on `remove`, and ensures path exists on `add` operations for JSON Patch. 
Ref: https://github.com/evanphx/json-patch/issues/138

Fixes #1249 

```release-note

```

```docs

```

### DIFF
--- a/pkg/modules/generators/app_configurations_generator.go
+++ b/pkg/modules/generators/app_configurations_generator.go
@@ -232,7 +232,13 @@ func JSONPatch(resources v1.Resources, patcher *v1.Patcher) error {
 					return fmt.Errorf("decode json patch:%s failed with error %w", jsonPatcher.Payload, err)
 				}
 
-				modified, err := patch.Apply([]byte(target))
+				// Apply JSON Patch with options to allow missing path on `remove`,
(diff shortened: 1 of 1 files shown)


Liuxingyu1111111 (approved):
lgtm

PR #1279, opened 2024-08-22
fix: bugs of watching while apply

### LINKED ISSUE #1273: Unexpected release result when receiving termination signal during `kusion apply` process
### What happened?

I had to stop the `kusion apply` process due to an `ImagePullErr` error that appeared during deployment.  When I tried the `kusion apply` command again, it was showing `all resources are reconciled. No diff found.` The `kusion destroy` was also not working in this case with `No managed resources to destroy` indicated. I can no longer manage the resource deployment using `kusion` in this case. The code below is from the release file, and although `spec` is empty, `state` is shown as succeeded.
```yaml
project: quickstart
workspace: dev
revision: 6
stack: default
spec:
    resources: []
    secretStore:
        provider: null
    context: {}
state:
    resources: []
phase: succeeded
```

Receiving a termination signal during `kusion apply` should be treated the same as normal apply failure.

1. Use `kusion apply` as usual.
2. Terminate this process halfway.
3. Retry `kusion apply` and `kusion destroy`.

_No response_

### Kusion version

<details>

```console
$ kusion version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

### Install tools

<details>

</details>

liu-hm19:
#### What type of PR is this?

/kind bug
This PR fixes some bugs of watching while executing the `kusion apply` command. 

Fixes #1273 

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/apply/apply.go
+++ b/pkg/cmd/apply/apply.go
@@ -419,8 +419,10 @@ func (o *ApplyOptions) run(rel *apiv1.Release, storage release.Storage) (err err
 
 	// start applying
 	fmt.Printf("\nStart applying diffs ...\n")
-	var updatedRel *apiv1.Release
-	updatedRel, err = Apply(o, storage, rel, changes)
+
+	// NOTE: release should be updated in the process of apply, so as to avoid the problem
+	// of being unable to update after being terminated by SIGINT or SIGTERM.
+	_, err = Apply(o, storage, rel, changes)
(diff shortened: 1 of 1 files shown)


Yangyang96 (approved):
LGTM

PR #1286, opened 2024-09-10
feat: add resource graph

Yangyang96:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/engine/runtime/kubernetes/kubernetes_runtime.go
+++ b/pkg/engine/runtime/kubernetes/kubernetes_runtime.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	yamlv2 "gopkg.in/yaml.v2"
@@ -44,6 +45,12 @@ var _ runtime.Runtime = (*KubernetesRuntime)(nil)
 type KubernetesRuntime struct {
 	client dynamic.Interface
 	mapper meta.RESTMapper
+}
+
+// KubernetesWatchEvent is a wrapper of k8swatch.Event
+type KubernetesWatchEvent struct {
+	Event    <-chan k8swatch.Event
+	Resource *unstructured.Unstructured
 }
 
 // NewKubernetesRuntime create a new KubernetesRuntime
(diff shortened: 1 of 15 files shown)


pkg/engine/runtime/kubernetes/kubernetes_runtime.go
```diff
@@ -724,6 +764,31 @@ func getNamedGVK(gvk schema.GroupVersionKind) schema.GroupVersionKind {
 	}
 }
 
+// Get the amount of dependent resources of the owner
+func getDependentAmount(object *unstructured.Unstructured) int {
```
liu-hm19, line 768: For the default amount of `Deployment`, `StatefulSet` and `CronJob`, maybe `1` is better than `-1` because there should be only one pod owned by these resources if `replicas` or `jobTemplate` are not declared.

pkg/engine/runtime/kubernetes/kubernetes_runtime.go
```diff
@@ -543,68 +574,51 @@ func convertString2Unstructured(yamlContent []byte) (*unstructured.Unstructured,
...
-	// Wait for the first watched obj
-	first := true
+// doWatch send KubernetesWatchEvent Slice which contains all the watch channels and their belonging resources if check ok
+func doWatch(ctx context.Context, watcher k8swatch.Interface, amount int, checker func(watched *unstructured.Unstructured) bool) []KubernetesWatchEvent {
+	events := []KubernetesWatchEvent{}
+	eventMap := make(map[string]chan k8swatch.Event)
 	signal := make(chan struct{})
+	timeoutCh := make(chan struct{})
+	if amount == -1 {
+		go func() {
```
liu-hm19, line 615: Kusion may miss the K8s event if the api-server processes longer than 2 seconds, I think it might be better to wait for the timeout after receiving the first event.

pkg/engine/runtime/kubernetes/kubernetes_runtime.go
Yangyang96: > For the default amount of `Deployment`, `StatefulSet` and `CronJob`, maybe `1` is better than `-1` because there should be only one pod owned by these resources if `replicas` or `jobTemplate` are not declared.
Good catch. Thanks!

pkg/engine/runtime/kubernetes/kubernetes_runtime.go
Yangyang96: > Kusion may miss the K8s event if the api-server processes longer than 2 seconds, I think it might be better to wait for the timeout after receiving the first event.

Will fix.

PR #1299, opened 2024-10-01
feat: add resource show command

### LINKED ISSUE #1261: Feat: kusion resource command

Command `kusion resource` to interact with resources :
- [ ] ~Add new command `kusion resource list` to show all resources of a specific project in a workspace.~
      - `--workspace`: optional, default is current workspace
      - `--project`: required
      - `--backend`: optional, default is current backend
- [ ] Add new command `kusion resource show {id}` to show the latest state of the specific resource by the id.
      - `--workspace`: optional, default is current workspace
      - `--project`: required
      - `--backend`: optional, default is current backend
### Why is this needed?
- It serves to check the status and get resource information by ID, by project, by workspace, or by all resources.
- To get the information about resources of a project that are applied in a workspace, from that we can get the `kusion-id` and the terraform import id of these resources and use them for another project in the same workspace.

hoangndst:
#### What type of PR is this?

/kind feature
Show details of a resource of the current or specified stack.

Fixes #1261

```release-note

```

```docs

```

### DIFF
--- a/pkg/cmd/resource/show.go
+++ b/pkg/cmd/resource/show.go
@@ -198,13 +198,11 @@ func (o *ShowOptions) Validate(cmd *cobra.Command, args []string) error {
 func (o *ShowOptions) Run() (err error) {
 	rel, err := o.ReleaseStorage.Get(o.ReleaseStorage.GetLatestRevision())
 	if err != nil {
-		return fmt.Errorf("no release found for project: %s, workspace: %s\n",
-			*o.Project, *o.Workspace)
+		return fmt.Errorf("no release found for project: %s, workspace: %s", *o.Project, *o.Workspace)
 	}
 
 	if rel.Spec.Resources == nil || len(rel.Spec.Resources) == 0 {
-		return fmt.Errorf("no resources found for project: %s, workspace: %s\n",
-			*o.Project, *o.Workspace)
+		return fmt.Errorf("no resources found for project: %s, workspace: %s", *o.Project, *o.Workspace)
 	}
 	resourceMap := make(map[string]apiv1.Resource)
 	for _, res := range rel.Spec.Resources {
--- a/pkg/cmd/resource/show.go
+++ b/pkg/cmd/resource/show.go
@@ -1,0 +1,255 @@
+package resource
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/liu-hm19/pterm"
+
+	"gopkg.in/yaml.v3"
+
+	"k8s.io/kubectl/pkg/util/templates"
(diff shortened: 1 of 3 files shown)


hoangndst:
cc @Yangyang96 @ffforest @liu-hm19

pkg/cmd/resource/show.go
```diff
@@ -0,0 +1,253 @@
...
+	for _, res := range rel.Spec.Resources {
+		resourceMap[res.ResourceKey()] = res
+	}
+
+	stateMap := make(map[string]apiv1.Resource)
+	for _, res := range rel.State.Resources {
+		stateMap[res.ResourceKey()] = res
+	}
+
+	if _, ok := resourceMap[*o.ID]; !ok {
```
liu-hm19, line 217: Is it necessary to consider a special corner case here, where a specified resource in the `Release` file is not in the `Spec` but still exists in the `State`? For example, the last `kusion destroy` failed. And if we consider this situation, we might not be able to return an error directly here. cc @Yangyang96

liu-hm19 (approved):

pkg/cmd/resource/show.go
Yangyang96: > Is it necessary to consider a special corner case here, where a specified resource in the `Release` file is not in the `Spec` but still exists in the `State`? For example, the last `kusion destroy` failed. And if we consider this situation, we might not be able to return an error directly here. cc @Yangyang96

After discussing with @liu-hm19, we have decided to leave the code as it is and set aside this particular corner case. The example above did not sufficiently explain this situation. This case arises when a user attempts to remove a resource using the `kusion apply` operation but it fails, while other commands such as `kusion resource graph` and `kusion destroy` do not handle it differently. This limitation is imposed by the structure of the command-line itself, and `kusion` will not be able to address all of these corner cases. As a result, we have decided to retain your implementation as it accurately reflects the initial user intentions. @hoangndst It's your decision to make to leave it here or change your implementation.

Yangyang96 (approved):

pkg/cmd/resource/show.go
hoangndst: For the original purpose of `kusion resource show`, in my opinion, we should keep this approach. 
Thank you for taking the time to provide your comments. 
cc @liu-hm19 @Yangyang96

liu-hm19: OK 👍 We agree with you. 

And during the discussion about the corner case, we've found a bug concerning the resource elimination in the `State` of the `Release` file when executing the `kusion destroy` command 😂 We will open an issue to track and fix this bug in the future.

PR #1363, opened 2024-12-26
fix: the problem of dependency change status not being updated

### LINKED ISSUE #1358: Bug: DependsOn in state does not update accordingly if resource is unchanged
### What happened?

Pre-requisite for this bug:
- resource A depends on resource B
- resource B name changed -> i.e. resource B ID changed
- resource A stays unchanged in the last apply

State before apply:
```
xxx:xxx:xxx:ResourceBOld:
  ...
xxx:xxx:xxx:ResourceA:
  ...
  DependsOn:
  - xxx:xxx:xxx:ResourceBOld
```

State after a successful apply:
```
xxx:xxx:xxx:**ResourceBNew**:
  ...
xxx:xxx:xxx:ResourceA:
  ...
  DependsOn:
  - xxx:xxx:xxx:**ResourceBOld**
```

This will cause an issue in the next apply:
`can't find resource by key:xxx:xxx:xxx:ResourceBOld in models or state`

After a successful apply:
State:
```
xxx:xxx:xxx:ResourceBNew:
  ...
xxx:xxx:xxx:ResourceA:
  ...
  DependsOn:
  - xxx:xxx:xxx:ResourceB**New**
```

All of the following conditions are met:
1. resource A depends on resource B
2. resource B name changed -> i.e. resource B ID changed
3. resource A stays unchanged in the last apply

In case of an unchanged resource, `res` is set to `prior`, aka. the resource in the last release state:
https://github.com/KusionStack/kusion/blob/main/pkg/engine/operation/graph/resource_node.go#L349

`res` gets updated as part of `RefreshResourceIndex`:
https://github.com/KusionStack/kusion/blob/main/pkg/engine/operation/graph/resource_node.go#L360

and later gets updated into the new release:
https://github.com/KusionStack/kusion/blob/main/pkg/engine/operation/graph/resource_node.go#L363

### Kusion version

<details>

```console
$ kusion version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

### Install tools

<details>

</details>

riven-blade:
#### What type of PR is this?

bug
Resource dependency status maintenance and improvement needs

Fixes #1358

```release-note
Solve the problem that the resources have not changed but the dependencies have changed, but the resource dependencies in status maintenance have not changed.

```

```docs

```

### DIFF
--- a/pkg/engine/operation/graph/resource_node.go
+++ b/pkg/engine/operation/graph/resource_node.go
@@ -347,6 +347,7 @@ func (rn *ResourceNode) applyResource(operation *models.Operation, prior, planed
 			log.Debugf("import resource:%s, resource:%v", planed.ID, json.Marshal2String(s))
 			res = response.Resource
 		} else {
+			prior.DependsOn = planed.DependsOn
 			res = prior
 		}
 	default:
--- a/pkg/engine/operation/graph/resource_node.go
+++ b/pkg/engine/operation/graph/resource_node.go
@@ -347,8 +347,12 @@ func (rn *ResourceNode) applyResource(operation *models.Operation, prior, planed
 			log.Debugf("import resource:%s, resource:%v", planed.ID, json.Marshal2String(s))
 			res = response.Resource
 		} else {
-			prior.DependsOn = planed.DependsOn
-			res = prior
+			copyPrior, e := prior.DeepCopy()
+			if e != nil {
+				return v1.NewErrorStatus(e)
+			}
+			res = copyPrior
+			res.DependsOn = planed.DependsOn
 		}
 	default:
 		return v1.NewErrorStatus(fmt.Errorf("unknown action type: %v", rn.Action))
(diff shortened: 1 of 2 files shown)


pkg/engine/operation/graph/resource_node.go
ffforest, line 1: Hi @riven-blade, although this looks like it may fix the issue, the value of `prior` is changed. Since `prior` is a pointer a `v1.Resource` that lives outside the scope of the current function, this can cause unexpected results outside of the function call in the caller of `applyResource()` (inside `func (rn *ResourceNode) Execute(operation *models.Operation)`).
Even though there are no logic after the call to `applyResource()` currently, this is considered to be a bad practice because it may cause problems when we add additional logic later.
I think it'll be better to just update `res.DependsOn` since `res` is scoped to the `applyResource()` only.

riven-blade: Yes, I agree with you. I have made adjustments to the code here. Take a look.

ffforest (approved):
lgtm

PR #1370, opened 2025-01-03
feat: support release rollback

### LINKED ISSUE #1028: feat: support roll-back, i.e. kusion apply with specified release

support roll-back, i.e. kusion apply with specified release

### Why is this needed?

support roll-back, i.e. kusion apply with specified release

riven-blade:
#### What type of PR is this?

feature

feat: rollback release

Fixes #1028 

```release-note
feat: support release rollback
```

```docs
feat: support release rollback
```


liu-hm19:
Thanks for your contribution! We'll review this PR a.s.a.p : )

liu-hm19:
@riven-blade Hi, I've just reviewed the code you submitted, and I think that it looks good to me overall. However, there are still two points I'd like to confirm: 
- For the `kusion release rollback` command, should it only allow users to roll back to a previously successfully applied release? Since rolling back to a failed revision doesn't seem meaningful, should we add some validation for this in the command? 
- Why can't we directly reuse the existing `Apply` and `State` related methods? Instead, we have to implement them again, especially creating a new `State` struct in the `release` package. Could you please clarify the reasoning behind this?

riven-blade:
Hi, I have made major adjustments to the abstraction of the previous apply logic, and the different operations between apply and rollback are completed in the prepare stage.  At the same time, the place where fixme was identified in the previous apply logic has been fixed.   Now there is no problem with the code logic when I test it, but I think there are still areas where the code can be optimized, and the priority is not that high. I don’t want a PR to be so heavy

